PR #75, opened 2014-12-22
Ward/better text editor

WardCunningham:
Oops. Forgot to fill this out. See below.


WardCunningham:
This pull request resolves the lost data issue identified in https://github.com/WardCunningham/Smallest-Federated-Wiki/issues/337. Splitting on the zeroth character position does not open a new editor above the current one but it does not lose any data either.

PR #81, opened 2015-01-04
try harder to find a reference slug

WardCunningham:
I've seen reference items send users to welcome-visitors. This is the default if the reference doesn't have a slug for the referenced page. I don't know when this started happening but I'm seeing lots of references without slugs. Perhaps the revision to the drop logic let this one slip in.

I've added a fall-back to making the slug from the title if title is present. This restores the desired behavior if by an alternate means. I'm deep in another set of commits or I would track down the root cause.

### DIFF
--- a/lib/reference.coffee
+++ b/lib/reference.coffee
@@ -5,11 +5,14 @@
 
 editor = require './editor'
 resolve = require './resolve'
+page = require './page'
 
 # see http://fed.wiki.org/about-reference-plugin.html
 
 emit = ($item, item) ->
-  slug = item.slug or 'welcome-visitors'
+  slug = item.slug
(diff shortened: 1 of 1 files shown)


WardCunningham:
Here is how the item.slug use to be set before the drop refactoring.
https://github.com/fedwiki/wiki-client/blob/3a927ddd5612a7291ef7f80c055079c2e816fe2e/lib/factory.coffee#L119

One more assignment here will bring the item back into balance.
https://github.com/fedwiki/wiki-client/blob/master/lib/factory.coffee#L70-L73

I'm hesitant to make this change because I already have to pr's in flight.

paul90:
Published as part of **+ wiki-client@0.4.2**

Issue #80, opened 2015-01-03
Need more thorough and correct html escaping


WardCunningham:
We introduced HTML escapes in items of type paragraph with pull request #71.

Since then we've found additional plugins that should perform similar escaping. Usually this involves rendering some form of caption or label.

We've also found that escaping interferes with proper formation of slugs used for internal links. Consider the link [[This & That]]. The #71 implementation links to this-amp-that because the ampersand is converted to &amp before the internal link is processed.

    $item.append "<p>#{resolve.resolveLinks(safe text)}</p>"

We can't resolve before escaping because generated html would be escaped. The fix requires coordinated resolving and escaping as a single operation. If resolveLinks were so improved it would seem to solve our caption problems mentioned above.

WardCunningham:
This issue reminds us that a page title can contain dangerous text and should be escaped before being rendered into html.

WardCunningham:
We have not seen any of these problems since #82.

PR #82, opened 2015-01-04
Improved html escapes now handled by resolveLinks

WardCunningham:
These commits improve html escapes as suggested in issue #80.
(Somehow I've picked up an independent commit appearing in a separate pull request.)

We combine resolving links with escaping html. The link resolution happens in two steps with the html escaping occurring between them. This is the comment from the relevant code.

```
  # markup conversion happens in four phases:
  #   - unexpected markers are adulterated
  #   - links are found, converted, and stashed away properly escaped
  #   - remaining text is sanitized and/or escaped
  #   - unique markers are replaced with unstashed links
```
The resolveLinks call now takes a second parameter where one might provide an alternative to simple html escaping. Alternative markup plugins will use this to offer their alternative translation method. The method provided must be secure in that it must effectively escape in addition to what ever else it does. There will be companion pull requests to these plugins that should be released together.

We also escape the page title text. This is step two of closing the xss vulnerability. We should audit every place we render any text from remote json and ensure appropriate escapes are in place.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -74,6 +74,7 @@ module.exports = function (grunt) {
           'test/lineup.coffee',
           'test/drop.coffee',
           'test/revision.coffee',
+          'test/resolve.coffee',
           'test/wiki.coffee'
         ]
       }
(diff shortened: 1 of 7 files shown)


paul90:
Published as part of **+ wiki-client@0.4.2**

PR #86, opened 2015-01-11
Ward/prompts

WardCunningham:
The factory has a little used prompt field which can guide users to enter the proper thing.
This commit makes this property easily editable. Shift-double-click to edit the prompt.
This commit also adds an option to the textEditor that allows one to specify what field is being edited.


paul90:
Published as **+ wiki-client@0.4.3**

PR #87, opened 2015-01-12
save true item.id as item.alias to avoid duplicates

WardCunningham:
This commit adds logic to the item drop handler to assign a new id to any item that conflicts with any existing item with the same id. This conflict happens when dragging multiple copies of an item into a single page.

This can happen when pulling items out of history. You can keep recalling the item from history and keep pulling the recalled item on to the same page. Move depends on unique ids and probably other things too.

The original id will be restored if the aliased item is subsequently dragged to a page without id conflict.

Journal merge will have to deal with similar aliasing if it is to be reliable.

Items that have been aliased store their original id in a field named 'alias'. Yes, I know this is counter intuitive.

### DIFF
--- a/lib/page.coffee
+++ b/lib/page.coffee
@@ -92,6 +92,11 @@ newPage = (json, site) ->
     item = _.extend {}, {id: random.itemId()}, item
     page.story.push item
 
+  getItem = (id) ->
+    for item in page.story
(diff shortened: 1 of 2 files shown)


paul90:
Not sure how I missed this one, published as **+ wiki-client@0.4.4**

PR #89, opened 2015-01-19
Improved put logic makes editing in localStorage viable.

WardCunningham:
This pull request replaces the hopelessly convoluted incremental pageHandler.put logic that is engaged when a pages is automatically forked to local storage. This uses the newish pageObject as the trusted representation to be saved (the product of our core client refactoring) along with emitting a 'refresh' event that communicates from the bowels of storage to the user interface, the first that we've employed this long-sought approach.

With this we need not fear the "yellow halo of doom" as it has become known.

The previously serious deficiency is now replaced with several known inconveniences.
* The textEditor split logic is foiled by the 'refresh' and reverts to the same behavior as other item types.
* A shared memory bug causes some history loss in the journal with repeated edits.

We include a debugging utility that writes a summary of all four page representations to console.log upon a shift-double-click. These are dom, model, localStorage and server.

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -149,7 +149,6 @@ $ ->
       pageObject = lineup.atKey $page.data('key')
       action = {type: 'fork'}
       if $page.hasClass('local')
-        return if pageHandler.useLocalStorage()
         $page.removeClass('local')
       else if pageObject.isRemote()
         action.site = pageObject.getRemoteSite()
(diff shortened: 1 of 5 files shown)


WardCunningham:
Note on terminology:

By inconvenience I mean a serious bug who's existence would be sufficient cause to delay a merge if it weren't replacing code know to lead to data loss.

WardCunningham:
I demo'd this behavior on the hangout this morning and was surprised that it didn't behave as expected. Here is what changed:

* I had been testing local editing by logging out and then continuing editing.
* Today, in the demo, I was editing a local storage page but with the server logged in.

The save logic in this pull request incorrectly simplified the case analysis on put such that in the second case it was writing to the server even though it was editing a local page. This is why I was not getting the expected refresh events. The server writes don't (yet) emit refresh events.

I will correct this logic and have a subsequent pull request before I sleep tonight.

PR #93, opened 2015-02-05
a single tab per origin

paul90:
This is a start of what we talked about in the hangout (Feb. 4th 2015).

Links and events that that would result in a change of origin will open in a new tab. We use the site name to name the tabs, so we can reuse tabs.

Works well in Chrome (and Vivaldi), but in Firefox the initial tab does not get reused correctly - suspect this is a browser feature.

While focus is switched, correctly in Chrome, it does not work correctly in Firefox if it configured to open new windows in a new tab.

### DIFF
--- a/client.coffee
+++ b/client.coffee
@@ -1,5 +1,7 @@
+console.log "Window Name: " + window.name
+window.name = window.location.host
+
 window.wiki = require('./lib/wiki')
 require('./lib/legacy')
 require './lib/bind'
 require './lib/plugins'
-
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -23,6 +23,6 @@ module.exports = ($journal, action) ->
     $action
       .css("background-image", "url(//#{action.site}/favicon.png)")
       .attr("href", "//#{action.site}/#{$page.attr('id')}.html")
+      .attr("target", "#{action.site}")
       .data("site", action.site)
       .data("slug", $page.attr('id'))
-
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -144,6 +144,7 @@ emitHeader = ($header, $page, pageObject) ->
   $header.append """
     <h1 title="#{tooltip}">
       <a href="#{pageObject.siteLineup()}">
+      <a href="#{pageObject.siteLineup()}" target="#{remote}">
         <img src="//#{remote}/favicon.png" height="32px" class="favicon">
       </a> #{resolve.escape pageObject.getTitle()}
     </h1>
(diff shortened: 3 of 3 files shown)


paul90:
I think this is as good as it can get - can't see a way to improve the Firefox experience :disappointed: 

I don't think I have missed any links that are part of the client, but...

WardCunningham:
I like this a lot.

This will take some time before we reach the full benefit. The server you're on must have the feature, of course. But the server your going to must also have it in order to get the one-tab behavior, not just a new-tab behavior.

We're also only controlling the links we generate. Markup in an html plugin would likely not follow our target convention, nor should we expect it, I suppose.

paul90:
Published as **+ wiki-client@0.4.5**

judell:
Thanks! This should be really helpful.

almereyda:
@paul90 As you said, this code can be improved.

Pasting a monologue of me from `#fedwiki`.

---

Good day everyone.
I'm trying to fix a minor issue with integrating fedwiki in a [hackfoldr](https://github.com/allmende/hackfoldr-2.0/) mashup.

The issue:

1. When opening http://hack.transforlab.net and clicking on any despite the "TransforLab" link (depending on your resolution, you have to unfold the folder list) opens everything in the desired iframe.
2. After opening the Federated Wiki, all links, also from the `hackfoldr` navigation, get routed to new tabs/a new tab.

The questions:
1. How can wiki escape the iframe and alter the original application's behaviour?
2. How do I stop it from doing so?

Any help appreciated.

The issue seems to reside around https://github.com/fedwiki/wiki-client/pull/93/files and https://github.com/fedwiki/wiki-client/blob/d4bd7e8adb16c664187760308b0b3825070b6554/lib/refresh.coffee , if I'm not mistaken.

Somehow https://github.com/fedwiki/wiki-client/blob/d4bd7e8adb16c664187760308b0b3825070b6554/lib/refresh.coffee#L154 is it. Don't know how to circumvent it altering the parent frame context, too.

Only blocking scripts via a sandboxed iframe helps here, but then I also lose the dynamic lineup. http://www.html5rocks.com/en/tutorials/security/sandboxed-iframes/

---

I am helping myself now by opening wiki links in a new tab themselves only, so they don't override the `hackfoldr` link behaviour. But it would be nice to be able to integrate it in such a meta navigation as hackfoldr offers.

paul90:
@almereyda I was referring to a browser restriction that restricts how tabs get reused.

A bit difficult to comment on the problem you were seeing, as you have changed your code. If you have a problem with how the client is behaving please raise a separate issue.

PR #94, opened 2015-02-07
Shift for hover effects on items and actions.

WardCunningham:
This pull request improves the item/action hover logic several ways.

* The hover highlight and scroll effects require that the shift key be pressed.
* Shift will highlight immediately when already hovering.
* Highlight state (class .target) is set and cleared rather than toggled avoiding sync problems.
* Hover over events highlights and scrolls other pages.

The scroll positioning would be better if it aligned like items on item hover. This has many boundary conditions that I will explore and then submit in a later pull request.

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -11,6 +11,7 @@ lineup = require './lineup'
 drop = require './drop'
 dialog = require './dialog'
 link = require './link'
+target = require './target'
 
 asSlug = require('./page').asSlug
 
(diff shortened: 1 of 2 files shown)


WardCunningham:
I've done a preview install of these commits on fed.wiki.org. I've been using that code to explore some of the conversation-club views from december's happening. [lineup](http://fed.wiki.org/journal.hapgood.net/view-between-rooms/kate.au.fedwikihappening.net/view-between-rooms/frances.uk.fedwikihappening.net/view-between-rooms/forage.ward.fed.wiki.org/view-between-rooms)

WardCunningham:
This pull request is now complete and ready to be merged.

WardCunningham:
Ok, I drew a picture. Now this is ready to merge.

image

paul90:
Published as **+ wiki-client@0.4.6**

PR #98, opened 2015-02-20
Update style.css

WardCunningham:
Add room to bottom of wiki page to make end of scrolling obvious. This is especially useful when shift-hovering for alignment. See example (from internal log inspector).

screen shot 2015-02-20 at 2 59 23 pm

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -243,7 +243,9 @@ p.readout {
 
 .footer {
   clear: both;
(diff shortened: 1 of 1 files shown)


WardCunningham:
I should acknowledge that any insight into css here comes from working with @1337807 Jonan Scheffler.

PR #99, opened 2015-02-21
License update

paul90:


pombredanne:
@WardCunningham @paul90 : Thank you my lords! This makes it easy now for others to consume and enjoy your wares :+1:

PR #102, opened 2015-03-01
tolerate nulls for items and actions

WardCunningham:
We've seen trouble rendering items cascade backwards through successfully rendered items. We're traced this to unexpected null items in a story. This pull request makes rendering of both items and actions more tolerant of nulls by carrying on as best as can be expected.

image

### DIFF
--- a/lib/page.coffee
+++ b/lib/page.coffee
@@ -39,7 +39,7 @@ newPage = (json, site) ->
     context = ['view']
     context.push site if isRemote()
     addContext = (site) -> context.push site if site? and not _.include context, site
-    addContext action.site for action in page.journal.slice(0).reverse()
+    addContext action?.site for action in page.journal.slice(0).reverse()
     context
 
   isPlugin = ->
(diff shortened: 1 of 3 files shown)


WardCunningham:
Alyson Indrunas helped me understand and track down this issue by assembling this screencast and then walking through this and several other compounding issues while screen sharing.

http://screencast.com/t/JfC8uiN8JJ (crazy error compounding at 1:30 in video)

paul90:
Do we have a way of reliably recreating this problem? It would be good to get to the root cause, but meanwhile it is good to handle the problems it causes.

I also notice looking at Alyson's site that a number of paragraphs appear to only contain spaces.

WardCunningham:
Regarding unexpected spaces: Alyson tells me that she would cut and paste content from other sources and that the result often needed cleaning up before it looked right. If one copies content from html pages some browsers attempt to preserve formatting for drop targets that can use it. I'm vague on how all this works. But I'm assuming that the clean up she describes comes from that. (The text editor handles newlines differently on paste vs. direct entry.)

Regarding unexpected nulls: I now suspect a programming bug on my part triggered the null sensitivity trigger a cascade of errors I saw in my own work. I have been careless synchronizing log writing and log reading in a custom server environment I was debugging. The json I produced from those logs could have had nulls. I've yet to catch it red handed.

Regarding tests for these commits: I created long pages with many similar paragraphs. My first thought was that there was some problem refreshing these pages especially when multiply viewed. This all worked fine. I went forward splicing in missing fields and eventually missing objects by editing the stored json with a text editor. I would like to capture each of these cases in unit tests but the current bugs are close to dom manipulations and thus harder to test.

PR #105, opened 2015-03-22
exposing `wiki.neighborhoodObject`

paul90:
so we can reuse search within the activity plugin - see fedwiki/wiki-plugin-activity#8


WardCunningham:
I don't know what I was thinking when I chose to expose only a portion of the neighborhood module through the wiki global. So now when we need the rest we must introduce a second name. I'm happy with the awkward wiki.neighborhoodObject because it makes relatively clear what happened rather than trying to propagate a breaking revision of the original global.

paul90:
Published as **+ wiki-client@0.4.9**

PR #108, opened 2015-03-28
correct page flag's href after fork

WardCunningham:
The pageHandler rewrites the dom of a page when forked so that it appears to have come from the origin. This commit revises this logic to make an acceptable href for the page's flag.

This commit addresses the issue raised in https://github.com/fedwiki/wiki-node/issues/44 where a recently forked page could not be dropped on a Factory with expected result.


paul90:
Published as **+ wiki-client@0.4.10**

PR #112, opened 2015-04-10
html formatted error message now requires html plugin

WardCunningham:
This corrects a formatting problem exposed by https://github.com/fedwiki/wiki-node/issues/45.
Formatting could be improved but this at least corrects the regression.


paul90:
Published as **+ wiki-client@0.4.11**

PR #113, opened 2015-04-19
Add site name mouse over to forks in journal

nrn:
Paired with @WardCunningham to add the site location to the hover effect in the journal, for actions that have an associated site.


paul90:
Published as **+ wiki-client@0.4.12**

Issue #117, opened 2015-06-07
Resize large images on drop before upload


WardCunningham:
Experience shows that images with a width of around 500 pixels work well with our current image encoding. This size works even better when the image has been aggressively compressed to 25-50kb.

The server expects json that is 5mb or less and will choke on more so drops that url encode larger than this just fail. Fork expects to post whole pages as a single json. That means the sum of all images, including edits of image captions, but come in under 5mb.

This issue suggests that the Factory's client-side file drop handling resize large images to 500 pixels when they are substantially larger than this. Drop handling could further capture source urls for images dropped from the web making higher resolution versions available.

An alternative and more complicated approach to image assets has been suggested elsewhere.
http://ward.fed.wiki.org/image-assets.html

# Resources

Drop decoding in wiki-client.
https://github.com/fedwiki/wiki-client/blob/master/lib/drop.coffee

Current image handling in Factory plugin.
https://github.com/fedwiki/wiki-client/blob/8dcc6074050a488dba5f1148173e4f4471d189b1/lib/factory.coffee#L95-L101

Image to canvas conversion in the Efficiency plugin.
https://github.com/fedwiki/wiki-plugin-efficiency/blob/master/client/efficiency.coffee#L39-L53

Resizing Images in the Browser using an HTML Canvas.
http://www.benknowscode.com/2014/01/resizing-images-in-browser-using-canvas.html

jQuery Plugin for Client Side Image Resizing
http://www.jqueryscript.net/layout/jQuery-Plugin-for-Client-Side-Image-Resizing-canvasResize.html

WardCunningham:
Making progress on this based on the Patrick Oswald version of BenKnowsCode.com.

Testing with this 3.5mb image from commons.
http://commons.wikimedia.org/wiki/File:Dry_Falls.jpg

The squeeze takes 700ms. The total time from drop to post response is only twice that.

We'll now try dropping that image on this comment. The github drop handling completed in about 5 seconds showing only markup. Preview took another 2 seconds to render from their server with subsequent views much faster. To their credit this is at high or full resolution. Something to aspire to.

![dry_falls](https://cloud.githubusercontent.com/assets/12127/8025834/88fd5388-0d18-11e5-8f6b-2df02ca15110.jpg)

WardCunningham:
[...]

PR #118, opened 2015-06-07
Resize and compress large images on drop.

WardCunningham:
This pull request is for the feature discussed in issue #117.

We stop short of reading or revising EXIF metadata. This leads to some drops appearing sideways.


paul90:
Published as part of + wiki-client@0.4.13

PR #119, opened 2015-06-10
shift-double-click for html or markdown

WardCunningham:
Wiki shortens slightly the path to making Paragraphs over other item types. This pull requests extends slightly the shift-double-click hack that converts Paragraphs to the now sanitized HTML type by considering alternatives.

Shift-double-click now inspects the Paragraph for the presence of various tags before converting to HTML. If absent, the converted item will be type Markdown instead. This is the list of expected tags extracted from the regular expression used to perform the test.

```
(i|b|p|a|h\d|hr|br|li|img|div|span|table|blockquote)
```

### DIFF
--- a/lib/paragraph.coffee
+++ b/lib/paragraph.coffee
@@ -6,6 +6,12 @@ editor = require './editor'
 resolve = require './resolve'
 itemz = require './itemz'
 
+type = (text) ->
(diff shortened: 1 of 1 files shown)


paul90:
Published as part of + wiki-client@0.4.13

PR #120, opened 2015-06-15
Drag and drop to import from export files

WardCunningham:
The server will bundle all site pages into an export file easily downloaded into a single file. [json](http://divergence.hapgood.net/system/export.json)

This pull request implements a simple way to recover individual pages from such a backup.

* Users export their sites to their own computer for backup.
* To recover a page, simply drag the backup file into the wiki tab.
* A ghost page opens listing the contents of the backup.
* Click any title to see the backup version as another ghost page.
* If it is the wanted page, fork it over your mistake.

This is what it might look like while recovering an accidentally damaged welcome page. Notice that both pages are ghostly because they do not yet exist in the site. The welcome page will turn white if forked.

image

We use the existing fork mechanism on the server to import files. This places an upper limit on the size of a page (5Mb) but not an upper limit on the size of a site. A future commit will add a convenient way to load all pages of an export into a new empty site.

### DIFF
--- a/lib/importer.coffee
+++ b/lib/importer.coffee
@@ -17,8 +17,11 @@ emit = ($item, item) ->
     result = []
     for slug, page of pages
       line = "<a href=#{slug}>#{ page.title || slug }</a>"
-      if page.journal && (date = page.journal[page.journal.length - 1].date)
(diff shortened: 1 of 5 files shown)


paul90:
Updated client published as + wiki-client@0.4.14

PR #122, opened 2015-06-18
Updating Dependencies +

paul90:
**Part of a group of breaking changes - DO NOT MERGE YET**

Make the client more self contained - move views from wiki-node-server, also see fedwiki/wiki-node-server#99


paul90:
Published as + wiki-client@0.5.0

PR #127, opened 2015-07-27
Correct jQuery UI built

paul90:
This is a fix for fedwiki/wiki-plugin-reduce#7


WardCunningham:
Is this simply a version synchronization issue between jQuery and jQuery UI?

paul90:
No, there are two downloads - the quick download (that that I assumed included everything, but doesn't), and the custom download.

WardCunningham:
Ah. Thanks.

paul90:
Published as + wiki-client@0.5.1

PR #130, opened 2015-08-15
show how to add things to ipfs

jbenet:
### Note: this PR is not to merge, but to show how it works.
---

This commit is a very simple example of how to use ipfs to store
assets. We basically just read a stream of data and pass it to
ipfs.add(stream). The response includes the hashes of files added.

(Of course, actually hosting the server-side fed wiki code on ipfs
and doing everything client side is possible. This is just a small
bit to start :)

On backing up the content:

This commit shows how to add content to an ipfs node. the node
will also "pin" the content, which means the node will back it up
and serve it to other nodes. Pins can be removed with:

    ipfs pin rm -r <hash>

On the API Address:

The ipfsApi('localhost', 5001) can be changed to point to whatever
node can be written to. Note that for now, while ipfs is still
figuring out either (a) the way to do a writable HTTP gateway with
low permissions, and (b) the node-ipfs implementation, we need to
setup CORS to work correctly with the server. You should be able to
do:

    ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["localhost:3000", "localhost:5001"]'

(or whatever ports you're using for fedwiki and the ipfs api)
If it doesn't appear to work, you can just run:

    ipfs daemon --unrestricted-api

But please do not run ipfs nodes like this all the time. it exposes
the API to any user.

### DIFF
--- a/lib/factory.coffee
+++ b/lib/factory.coffee
@@ -10,6 +10,12 @@ editor = require './editor'
 synopsis = require './synopsis'
 drop = require './drop'
 
+# ipfs things
+fileStream = require 'filereader-stream'
+ipfsApi = require 'ipfs-api'
+# TODO: change api addr + ports from user config
+ipfs = ipfsApi '127.0.0.1', '5001'
+
 escape = (line) ->
   line
     .replace(/&/g, '&amp;')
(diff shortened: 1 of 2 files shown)


jbenet:
Also, to install ipfs: 

```
# globally
npm install -g go-ipfs
ipfs init
ipfs daemon

# locally
npm install go-ipfs
node_modules/.bin/ipfs init
node_modules/.bin/ipfs daemon
```

And, programmatically:

    npm install ipfsd-ctl

then see https://github.com/ipfs/node-ipfs-api/blob/77a700bd0070ed09912f4e590b174bb8a2c65387/test/test.js#L25-L36

jbenet:
@WardCunningham finally got around to finishing up what we started!! Turns out it was a bunch of fixing dependencies, and how we used them. But see how small it is to get going! :) Please let me know if anything's not a breeze.

WardCunningham:
Sweet. Thanks.

almereyda:
@WardCunningham @jbenet We ( @acorbi @opn + me ) will be looking into turning this PR into a Plugin. It's a nice exercise.

paul90:
I'm going to close this PR - I've included a link here from [Wiki on IPFS](http://goals.pod.rodwell.me/wiki-on-ipfs.html).

PR #131, opened 2015-08-20
speed up sitemap loading by 4x

WardCunningham:
This changes the delay between loading sitemaps to 1/2 second. Let's make this 4x improvement the norm. This builds on @paul90's months of improvements to sitemap caching and now activity configuration and rendering.


paul90:
Published as + wiki-client@0.5.2

PR #135, opened 2015-10-04
Ward/transport

WardCunningham:
Expose newPage and showResult that will allow a plugin to create and display new pages.

These changes were developed by @paul90 and myself while experimenting with the Transport plugin. It is my intention to finish this plugin development once this commit has been published.


paul90:
Published as + wiki-client@0.5.3

PR #136, opened 2015-10-12
Batalyx/slug after pipe

batalyx:
As in Wikipedia, internal link [[foo|bar]] results to slug 'bar' instead
of 'foo'.  This is a great help with languages like Finnish, where the
body of the word can change. Now several links can point to one page.

I've tested this only in a small scale. It seem's to work :-) But I've had all kinds of odd problems, when trying to integrate this. Installing npm from github has not worked, and recompiling either until using 'npm link' in web-client directory, installing grunt, etc. So, I'd be happy if someone with more experience with fedwiki & node development would try this out.

### DIFF
--- a/lib/resolve.coffee
+++ b/lib/resolve.coffee
@@ -40,6 +40,10 @@ resolve.resolveLinks = (string, sanitize=escape) ->
     slug = asSlug name
     stash """<a class="internal" href="/#{slug}.html" data-page-name="#{slug}" title="#{resolve.resolutionContext.join(' => ')}">#{escape name}</a>"""
 
+  internalWithPipe = (match, name, slug) ->
+    slug = asSlug if slug then slug else name # if someone enters something unsluggish as slug
+    stash """<a class="internal" href="/#{slug}.html" data-page-name="#{slug}" title="#{resolve.resolutionContext.join(' => ')}">#{escape name}</a>"""
+
   external = (match, href, protocol, rest) ->
     stash """<a class="external" target="_blank" href="#{href}" title="#{href}" rel="nofollow">#{escape rest} <img src="/images/external-link-ltr-icon.png"></a>"""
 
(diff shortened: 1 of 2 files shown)


WardCunningham:
Thank you for bringing this work to our attention.

Please help me understand the problem this solves for page titles in Finnish.

Understand also that our unicode support is weak in this area but we have discussed ways to improve it. My hope would be that a simpler markup than Wikipedia can be made to serve all users.

batalyx:
Thanks for considering this. I'll try to explain more.

For example, the word *plug-in* translates to *liitännäinen* in Finnish. That's the basic form. But if we translate *"About Plug-ins"*, it translates to *"Tietoa liitännäisistä"* (literally: *"information about plug-ins"*). Now the suffix transforms the ending of the word from *-nen* to *-sistä*. The beginning of the word, *liitännäi*, is not a proper Finnish word, so it's not acceptable as a slug or title.

Then we have words like *kissa* (a cat), which has a plural genetive form of *kissojen* (*_ of cats*), where the body of the word transforms *kissa* -> *kisso-*. Or words like *kenkä* (a shoe) that transforms to *kengät* (shoes). This time the beginning *ken* is a proper word, but it is an old form of *kuka* (*who*).

And of course for links with multiple words there are many more changes.

There's a possibilty to write a new page for each form of the word, that would only contain a link to the main pages, but that's not readable any more.

So, it would make it easier to write links, if one could provide the slug in the link expression. Wikipedia, for example, uses the form `[[ link text | slug ]]`, which I think would work for fedwiki too. The implementation seems to be quite easy, just look for the last bar character and use everything after that as a slug.

I hope this clarifies. I thought of using the form `[[visible words] slug]`, but as Wikipedia and a few other wikis already use the bar-form, I used it in the final implementation.

Mind you, I don't think this should be implemented too quickly. I'm giving a course in Software Architectures this fall, and I'm planning to use fedwiki for the material, so I hope I'll get some experience how well this slug/link-notation works.

I think I can't imagine any easier approach. It might be that I've not tried enough. Maybe translation tables, that change, for example, *kissojen* to *kissa*, but that sounds a bit too hard to use. It would make it easy to implement "use any instance of this word as a link" kind of feature, alas.

Unicode would not help here a lot, but it will help a lot more languages than this change will.

WardCunningham:
@batalyx, Thank you for explaining the various word forms in Finnish. English has tense and number which complicates consistent page naming in any wiki.

With the smaller paragraphs in federated wiki we see less use of links embedded in sentences. This is not the same solution you suggest but it does lessen the grammatical constraints on link text. We are more likely to add the phrase "See Blah Blah" at the end of a paragraph than try to work "Blah Blah" into a sentence. 

Consider the article [We Real Cool](https://en.wikipedia.org/wiki/We_Real_Cool) about the poem with the same title. This isn't grammatical in any normal sense but we allow the author some grammatical latitude and respect it in our references. We elevate titles of federated wiki pages in a similar way and live with the grammatical consequences.

Let's hold onto this pull request for a while. Let us know if this more relaxed linking style works for you in your federated wiki writing. I hope it does.

WardCunningham:
Having considered this pull request and the following discussion for some time I am going to respectfully decline this modification. Adherence to grammatical convention in use diminishes the role of a page title as a known and recognized work. I appreciate that there is tension here, but believe we have it right as is. Thanks again for your contribution.

PR #137, opened 2015-10-20
Let factory grow vertically with lots of plugins.

WardCunningham:
This small change in style.css will allow a factory to grow vertically when there are more than the normal number of plugins in any one category.

image


WardCunningham:
@paul90 I'd like to merge this small change so that it is not forgotten.

PR #138, opened 2015-11-08
Add a paper scrollable div inside the pages viewport

sillydeveloper:
- This adds a new div called "paper" that contains the scrollable viewport instead of the page.
- Adds a "asTitle" utility method to convert a slug back into a title.

@WardCunningham will check this out for a bit with the various plugins to ensure it doesn't break anything.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -50,7 +50,7 @@ module.exports = function (grunt) {
       // build the client that we will include in the package
       packageClient: {
         src: ['./client.coffee'],
-        dest: 'client/client.max.js',
+        dest: 'client/client.js',
         options: {
           transform: ['coffeeify'],
           browserifyOptions: {
@@ -88,7 +88,7 @@ module.exports = function (grunt) {
                   ' Licensed <%= _.pluck(pkg.licenses, "type").join(", ") %> */'
         },
         files: {
-          'client/client.js': ['client/client.max.js']
+          'client/client.js': ['client/client.js']
         }
       }
     },
--- a/client/style.css
+++ b/client/style.css
@@ -145,13 +145,14 @@ footer {
   width: 100%; }
 
 .journal {
-  width: 420px;
+  /*width: 420px;*/
   overflow-x: hidden;
   margin-top: 2px;
   clear: both;
   background-color: #eeeeee;
   overflow: auto;
-  padding: 3px; }
+  /*padding: 3px; */
+}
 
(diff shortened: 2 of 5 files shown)


WardCunningham:
Edits to `static.html` and `editor.coffee` must have slipped in from other editing.

sillydeveloper:
Finally pushed up the requested changes!

WardCunningham:
We produce client.js and client.max.js but use only client.js in production installs.

I'm not seeing a second commit. Did you push code to this branch this morning?

paul90:
Yes `client.max.js` is correct - this is then passed through uglify to give the final client. 

May seem wrong as many would use `client.js` and `client.min.js`, but at the time this was put in place the map file linked back to the coffeescript source. *I wonder when that got broken...*

paul90:
Really not sure what this pull request is trying to achieve?

> This adds a new div called "paper" that contains the scrollable viewport instead of the page.

I don't see, or recall, any reason why this extra level of nesting is necessary!

> Adds a "asTitle" utility method to convert a slug back into a title.

What is this for? Conversion of a title into a slug is a one way action - created using `name.replace(/\s/g, '-').replace(/[^A-Za-z0-9-]/g, '').toLowerCase()`

As far as I can see this method is not achieving anything. The correct title is getting used rather than that created by this `asTitle` method. For example, a page with the title `a - b` is getting correctly displayed despite getting converted to a slug of `a---b` and then a title of `a   b` by this method.

WardCunningham:
This separates the two kinds of page halo we use. One kind rests around the page showing that it is active (think arrow keys), the other rests within the page showing its type (think blue, green and yellow). The extra layer of div allows both to be present at the same time. Another subtle feature is that the type halo scrolls with the "paper" so that one never read text through the halo.

The title manipulation was to "guess" at a more likely title when the page is being fetched. This lingers when operating on a slow internet. I suggested we go so far as to guess that the page first paragraph will be that which we find in a sitemap. This would be replaced with the full page and proper title once it arrives.

I had pointed Andrew in these directions while I wrote and he proof-read my introduction to hacking wiki.
http://dev.asia.wiki.org/view/hacking-federated-wiki/view/render-pages

Much of this slips from my memory so careful review is appropriate.

paul90:
That rings a bell, it had completely slipped from my mind.

paul90:
Thank you Andrew, @sillydeveloper

Issue #66, opened 2014-12-06
#fedwiki icq channel does not exist on freenode.net? 


svdoever:
I have so many questions, but not sure where to ask them. According to http://fed.wiki.org/view/frequently-asked-questions there is an ICQ channel of freenode.net:

There is an IRC channel, #fedwiki on freenode.net.

Problem is: I can't find it. Does it still exist? Is it the place to ask questions?

PR #140, opened 2015-11-14
changed styling of reference-div from background-color to box-shadow …

i2p-lbt:
…(as external pages have) and added style for paragraphs with the reference-div (according change to remove inline style in wiki-plugin-reference) - addressing #66 - needs the according pull-request on wiki-plugin-request to remove the inline-style there which get's replaced by the new ".reference p" from this change

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -379,5 +379,12 @@ textarea {
 }
 
 .reference {
-  background-color: #f0f8ff;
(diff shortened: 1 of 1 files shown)


paul90:
@i2p-lbt  Thank you for this change, I have made a minor change to replace the box-shadow with a border on the left.

Issue #126, opened 2015-06-25
Shouldn't fork a Future page


WardCunningham:
Linking to a non-existent page creates a ghost page with a Future plugin offering to create the page. The add button is ignored in this context but apparently not the fork button. Somehow, was it a timing thing?, I managed to fork the ghost page into my site. This ill-behaved page reported the following json even after reloading the browser.

```
{
  "title": "Science of Concurrency",
  "story": [
    {
      "id": "bd83c3991e6566a9",
      "type": "future",
      "text": "We could not find this page.",
      "title": "Science of Concurrency"
    }
  ],
  "journal": [
    {
      "type": "fork",
      "date": 1435242917084
    }
  ]
}
```

This renders as a non-ghost page with a journal showing the fork. But it won't make good on its offer to create the page. The add and fork buttons should be ignored for this page to avoid this unwanted state.

paul90:
Closing as resolved by #142

PR #142, opened 2015-12-28
lose the ghost flag whenever page forked

WardCunningham:
We correct a confusing bug where transported pages appear as ghosts but do not lose the ghostly appearance when forked into the origin site. 

We move the ghost-class removal code used when forking from history so that it is applied in all forking cases. No fork ever produces a ghost page.

@opn brought this bug to my attention.


WardCunningham:
Also fixes #126

PR #145, opened 2016-01-29
move width from page to paper

### LINKED ISSUE #144: Narrowing of wiki pages causing problem
An change in the css, included in #138, reduces the wiki page width from 430 to 370. This causes problem for plugins, such as video, that expect the wider columns. There is also content, using the html plugin, that written specifically for the 430 width wiki-pages, so simply changing video plugin is not going to be the solution.

I am aware that some mobile clients have screens for which 430 is too wide, a different approach would be to detect the screen width and use a css transform to scale the contents.

This results in wiki pages gaining a horizontal scroll bar.

paul90:
Fixes #144 by moving the width definition from `.page` to `.paper`

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -266,7 +266,7 @@ p.readout {
   float: left;    /* inline-block preferred? */
   margin: 8px;
   /* padding: 0 31px; */
-  width: 430px;
+  /* width: 430px; */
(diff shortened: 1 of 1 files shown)


WardCunningham:
Thanks.

PR #146, opened 2016-01-31
set a minimum height on .paper

paul90:
prevents bottom box-shadow from appearing partway up page on short pages.

:blush: didn't spot the problem with short pages until after #145 was merged.

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -288,8 +288,9 @@ p.readout {
 }
 
 .paper {
-  width: 420px;
+  width: 430px;
   padding: 30px;
+  min-height: 100%;
 }
 
 .factory,


WardCunningham:
I'm still having trouble building new versions of wiki-client. Here is what I see when I build this branch, npm link and then launch my previously working version of wiki. I'm sorry that I'm not more help. 

image

paul90:
Looks as if the width and padding are missing from your `.paper` style - or maybe the `paper` `div` is missing.

That red halo is not the standard one, so could it be that you have a site stylesheet that is causing this?

WardCunningham:
I thought perhaps my experiment with status/theme was interfering as you suggest but this is not the case. I suspect that npm-link is failing me again so will work on testing this with a more complete and isolated build. I'm worried that the 100% is working against something Andrew and I did intentionally so I just want to see the effect on the screen. Forgive me for being so slow. I have five hours on a train today so I will dig in then.

PR #149, opened 2016-02-02
use .find(...).length - true if something is found

### LINKED ISSUE #147: Problem forking page
In e892a4f (part of #142), a check is added to refuse to fork a page that contains the future plugin. Unfortunately, the check added prevents any pages being forked.

paul90:
fixes #147

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -137,7 +137,7 @@ $ ->
 
     .delegate '.fork-page', 'click', (e) ->
       $page = $(e.target).parents('.page')
-      return if $page.find('.future')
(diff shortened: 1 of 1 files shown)


paul90:
no while this may fix #147 it also prevents historic version of a page from being forked...

WardCunningham:
Let's talk about how we can better handle this sort of case analysis with confidence. Perhaps our nascent  pageObject can be a point of contact for these decisions. (Of course we will have to make them reliably available first.)

WardCunningham:
Catching up with pull requests. Sorry for the delay.

PR #151, opened 2016-02-28
style .paper from classes on parent .page

WardCunningham:
The .local, .remote and .plugin classes are added to .page divs to add halos and sometimes control behavior. This commit reverses the decision to move this annotation to the recently introduced .page child div.paper which is the medium that scrolls.

Rather than move the annotation, we style the .paper using the css combinator: child selector, denoted by the greater-than arrow. For example, this selector will style the .paper of a .remote page.
```
.page.remote > .paper { ... }
```

Regretfully v0.7.3 was published with css changes that were incompatible with our current .page.local and .page.plugin machinery thus omitting halos in these cases. Here all halos are visible.

image

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -275,15 +275,15 @@ p.readout {
     box-shadow: 2px 1px 24px rgba(0, 0, 0, 0.4);
     z-index: 10; }
 
-.paper.plugin {
+.page.plugin > .paper {
   box-shadow: inset 0px 0px 40px 0px rgba(0,220,0,.5);
 }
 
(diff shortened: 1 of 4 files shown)


WardCunningham:
We've looked through wiki-client/lib for jquery code that reaches up to .parent() and expects to find .page rather than .paper. We found two.

* The JSON link at page bottom.
* The testEditor split/join keystroke handling.

This additional commit replaces this approach with the more robust search, `.parents('.page')`.

PR #152, opened 2016-03-01
Adding checks for vulnerabilities in libraries and node packages

paul90:
Adding 
* retire.js - to check for libraries with known security problems.
* nsp - to check for known vulnerabilities in node packages


paul90:

PR #153, opened 2016-03-01
Update javascript libraries, and replace code using depreciated/removed jQuery features

paul90:
Update to jQuery libraries, replacing migrate version with known problems.

Update d3 to current version, we last updated a year ago...

Remove libraries that are no longer needed
* Modernizr only used to support ancient Internet Explorer versions (no longer supported by Microsoft).
* SockJS does not appear to be used anywhere.

The use of a number of depreciated/removed jQuery features, this was being masked by having jQuery migrate. To see messages from the migrate code, modify `views/static.html` to use the unminified versions of jQuery Migrate. 

N.B. I don't see anywhere `.delegate '.score', 'hover', (e) ->` in `lib/legacy.coffee`  is used.

Still need to check plugins for use of depreciated/removed jQuery features


WardCunningham:
The .score delegate is created in wiki-plugin-rollup. This click handling should be the responsibility of the plugin. I'm happy to make that change. Also, rollup should probably be moved to the fedwiki incubator/infirmary as it exists mostly as an example of what is possible.

paul90:

PR #154, opened 2016-03-19
move width from .paper to .page

WardCunningham:
This commit corrects two problems observed since we've introduced .paper to carry the colored halo.

* Pages were slightly larger than the div that held them. This meant they could wobble side to side by a few pixels.
* Page width was unconstrained on page load until javascript rewrote the page with the .paper div added.

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -266,7 +266,7 @@ p.readout {
   float: left;    /* inline-block preferred? */
   margin: 8px;
   /* padding: 0 31px; */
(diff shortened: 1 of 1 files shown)


WardCunningham:
@nrn worked out the problem here by comparing dom structures between my current and his older site.

* http://ward.asia.wiki.org/
* http://nrn.io/

Thank you Nick.

nrn:
Glad I could help!

PR #155, opened 2016-03-20
Ward/transport proxy

WardCunningham:
This pull request adds functionality to the Future plugin that will offer to reinvoke a transporter if a link from a previously transported page can't be found in the current context.

image

http://ward.asia.wiki.org/explore-transport-proxy.html


paul90:
Published in + wiki-client@0.7.6

PR #156, opened 2016-03-21
simplify reference styling

WardCunningham:
I'd like to make a page of References look a little better, more like natural text. We started coloring them so that one would see the first line and subsequent lines as one thing. Let's just merge subsequent lines onto the end of the first line so it does look like one thing even if not given any color treatment. 

image


paul90:
Published in + wiki-client@0.7.6

PR #158, opened 2016-05-03
update local sitemap on each put

WardCunningham:
Search can't find pages that were just added to a site because the sitemap is out of date. This pull request updates the sitemap for the origin on each successful page put.

This inaccuracy is especially annoying for the Graph plugin when clicks create new pages without changing the color code, green should turn to blue. Here I added Random Page to the RSP graph and then clicked it. I was asked if I wanted to create the page. See that it has turned blue when I clicked the create button in the adjacent Future page.

image

### DIFF
--- a/lib/neighborhood.coffee
+++ b/lib/neighborhood.coffee
@@ -52,6 +52,23 @@ neighborhood.registerNeighbor = (site)->
   populateSiteInfoFor( site, neighborInfo )
   $('body').trigger 'new-neighbor', site
 
+neighborhood.updateSitemap = (pageObject)->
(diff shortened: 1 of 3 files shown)


WardCunningham:
Further testing shows that Recent Changes now updates with each change. However the logic that shows site flags doesn't expect sitemap-done events repeated for the same site. Hold this pull request until that logic can be appropriately revised.

image

WardCunningham:
@paul90 Please merge this unless you can think of any issue we left undone in our work together weeks ago. Thanks.

paul90:

PR #159, opened 2016-05-07
preventing the horizontal scroll to unexpected places on creating item from factory

paul90:
Ward and myself looked into this *old* bug, where when creating an item it ends up being scrolled off to one side.


paul90:

PR #160, opened 2016-05-22
size the action background gradient if it is other than expected

WardCunningham:
This addresses a long-standing limitation wherein journal action background gradients were assumed to be gradients and of the dimensions of generated gradients. (I waited for the avatar style backgrounds to go away and even created the Flagmatic plugin to offer some choice within federated wiki branding. But good content remains with poorly rendered backgrounds.)

Consider:
http://journal.hapgood.net/artwork-as-recipe.html

Before:
image

After:
image

Notice that even the generated gradients look better since they too are sized and rendered in full.

### DIFF
--- a/client/style.css
+++ b/client/style.css
@@ -160,6 +160,7 @@ footer {
 .action {
   font-size: 0.9em;
   background-color: #cccccc;
+  background-size: 24px 24px;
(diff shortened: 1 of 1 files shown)


paul90:

PR #163, opened 2016-06-15
UX improvements : mobile nav and UI

clemsos:
After discussion with @WardCunningham and @opn during the OSCEdays in Bilbao, I made some basic improvements to the wiki-client UI.

* *page header*: "close" button to close page
* *page header*: "pin" button to pin page 
* *page header*: "close all pages" button to close all unpinned pages  
* *page footer*: "history" button to toggle journal
* *navbar*: remove unused menu in the bottom nav bar 
* *mobile*: mobile-first css and media-queries (a page is displayed full-width on mobile)
* *mobile*: swipe left/right to navigate between pages on mobile

One important finding in the process is that event handling could be improved to handle easily mobile events and behaviors. Here are some examples 

* On mobile, the 'press' or 'tap' or 'doubletap' event should trigger the edit of an item (replacing double-click)
*  the drag & drop should start only after a 'press' or 'hold' event on mobile (the drag & drop is triggered on touch which makes the scrolling difficult)

I couldn't find an easy way to catch the jQuery events and replace it with mobile events. Right now, it looks like events are mostly grouped in ```legacy.coffee``` but there is also events in [different places](https://github.com/fedwiki/wiki-client/search?utf8=%E2%9C%93&q=click+path%3Alib&type=Code) in the code. It may be good to put all events in a single place to define a more consistent approach to handle them. No idea how much work it will take though...


opn:
Fantastic. I think pulling the events into one place will help a lot.

WardCunningham:
There is a lot going on here. Some screen shots would be helpful. Thanks.

paul90:
Thank you @clemsos for an early sight for this work in progress. Some initial thoughts and observations:

Rather than create a pin system with it's own set of behaviours wouldn't it be better to mimic the behaviours exhibited by application tabs, when pinning has a very specific meaning. This would present a set of behaviours that most, if not all, users will be aware of rather than requiring them to learn something very slightly different. A typical tab system will offer closing other tabs, and tabs to the right, as well as closing the current tab. Offering the same for wiki pages would leverage this existing user experience.

That aside, there needs to be some visual feedback when a wiki page is pinned. There also needs to be consistency of action when 'close all unpinned' is pressed. Currently it varies depending on if any pages have been pinned, or not.

The character used for "close all unpinned", the "All equal to" sign `≌` does not convey the meaning that the its use will perform. While the "pin" doesn't fix pleasantly with the other user interface items.
screen shot 2016-06-15 at 14 51 53


The display of the wiki page is broken, with the bottom of the page disappearing behind the foot.
screen shot 2016-06-15 at 14 01 41

Displaying the journal requires a button press, and then a scroll to bring it into view. When the button is pressed again the page jumps as the page length is reduced when it is removed.

WardCunningham:
I should mention, this pull request will get a lot of attention because these improvements have been on our to-do list for a long time. You can just assume we're excited about everything here and working hard to understand your approach in each case.

http://ward.asia.wiki.org/touch-support.html
image

WardCunningham:
I notice that this pull request adds hammer.js without removing touch-punch. Wouldn't these modules compete to interpret touch events? Can we use just one?

image

WardCunningham:
We need to distinguish between persistent operations on the wiki and transient operations on a view into the wiki federation. Let's avoid using a look-alike button for delete page-view from lineup that could be confused with history making buttons like fork-to-here and add-item-here.

Circle-X is commonly used.

image

I'm partial to the version in the lower-left of this list of images. It's light weight implies that it is unsubstantial in its effect.

WardCunningham:
Active and mobile behaviors interact in unintended ways.

I've seen various versions of this in chrome on my macbook air. I will describe one and include a screenshot. Repeat this as follows:

* open a welcome page in a wide browser
* squeeze width of browser until page expands to fill browser tab
* click a link to open a new page, second page appears where expected
* squeeze width of browser some more, second page drifts out of view to the left

image

paul90:
> I notice that this pull request adds hammer.js without removing touch-punch. Wouldn't these modules compete to interpret touch events?

I see a number of people having issues with drag and drop when using jQuery-UI with touch-punch and draggable (which we use under the covers in sortable) and hammer.js - looks to be an event propagation problem. I'm not sure if re-writing the drag and shift-drag code to use native HTML5 sidesteps this problem. In any case this code has to be revisited to solve the visual problems with shift-drag not leaving a visible copy of the dragged item behind. I rather use a native HTML5 solution rather than jQuery draggable, as walks us into other problems.

WardCunningham:
@clemsos We are excited to incorporate these changes into wiki. As day-to-day users we will surely find many small things that need to be adjusted. I suggest that we focus on the touch work first and follow it through to any changes with activity or other services that might be impacted before considering lineup manipulations. To this end, would you consider separating these commits into one, touch specific, pull request? We can then focus on testing and/or coding touch details before launching into more changes to the javascript core. Thanks for considering this.

clemsos:
thanks @WardCunningham  @paul90 @opn for the feedbacks. I just sat back at my computer after a crazy week with some time to look more closely at your answers.

I will split this PR into 2 : one for the pin system, and the other one for the mobile/responsive layout. Those are 2 different things and need to be addressed separately. Let's follow up on discussions in each of them

PR #167, opened 2016-06-20
UI : pin / close pages and toggle journal

clemsos:
This is a follow-up on discussions in #163  regarding a basic pin / close system to make the navigation easier in fedwiki. Basically, the initial idea is to mimic browser behavior with a system to pin and close all pages.  I added a few buttons in the top left of the page header that implement this feature.   
- _close_ button to close page
- _pin_ button to pin page 
- _close all pages_ button to close all unpinned pages  

selection_136

Follow-ups on @paul90 comments on #163  

> Rather than create a pin system with it's own set of behaviours wouldn't it be better to mimic the behaviours exhibited by application tabs, when pinning has a very specific meaning. This would present a set of behaviours that most, if not all, users will be aware of rather than requiring them to learn something very slightly different. A typical tab system will offer closing other tabs, and tabs to the right, as well as closing the current tab. Offering the same for wiki pages would leverage this existing user experience.

Agree, my intent is to copy browsers behaviors. We need a "close all pages on the right" system, but this should come with a way to reorder the pages (namely a drag'n drop, which makes it a bit more complicated now) . 

> That aside, there needs to be some visual feedback when a wiki page is pinned. There also needs to be consistency of action when 'close all unpinned' is pressed. Currently it varies depending on if any pages have been pinned, or not.

"Close all unpinned" will close all pages that are not pinned, except if none are pinned. At least 1 page (the first) is left open to prevent blank screen.

> The character used for "close all unpinned", the "All equal to" sign ≌ does not convey the meaning that its use will perform. While the "pin" doesn't fit pleasantly with the other user interface items.

yes that's right, it was mostly a placeholder. I have it changed to the "empty set" character but we may need to use custom icons for this, even though I like the idea of using Unicode characters as icons. For the same reason (using Unicode),  the pin character will look different on all computers. We can solve that by adding a webfont and using maybe a [unicode-range](https://developer.mozilla.org/fr/docs/Web/CSS/@font-face/unicode-range) to prevent the font for applying to the whole document.

---

Also, the sometimes high number of icons in the journal added much visual noise in the footer, so I added a simple way to toggle it with a _history_ button.

selection_137

> Displaying the journal requires a button press, and then a scroll to bring it into view. When the button is pressed again the page jumps as the page length is reduced when it is removed.

I added an autoscroll in 5dc2586 to make reading the journal easier after showing it. I don't think the jump is a problem, but we can also make sth about this.

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -146,7 +146,10 @@ $ ->
     .delegate '.toggle-journal', 'click', (e) ->
       e.preventDefault()
       $page = $(e.target).parents('.page')
-      $page.find(".journal").toggle()
+      $journal = $page.find(".journal")
+      $journal.toggle()
+      if $journal.is(':visible')
+       $page.animate({scrollTop: $journal.offset().top});
 
     .delegate '.fork-page', 'click', (e) ->
       $page = $(e.target).parents('.page')


WardCunningham:
Note: this contribution has been obsoleted by the drag behavior added a few years ago.

PR #168, opened 2016-06-20
Mobile UX, UI and events handling in wiki-client 

clemsos:
_NB: I don't think the initial/current state of this PR is anywhere close to be ready to merge. It is more intended to be a place to discuss about mobile and solve problems gradually._*

IMHO there is three important preliminary parts of the discussions to have before starting with the mobile implementation.
- UI : switch to mobile-first
- UX : define gestures and behaviors on mobile
- events : define a clear way to manage events

Read previous notes from @WardCunningham  http://ward.asia.wiki.org/view/touch-support . 
## UI mobile-first

CSS should be switched to a "[mobile-first](https://codemyviews.com/blog/mobilefirst)" paradigm. The idea is that instead of degrading design to support different devices (responsive) you start with a default mobile design and you use media queries to have improved display on bigger devices (mobile-first). This is a good practice and looks quite easy to achieve as wiki-client UI is very clean already. 
- [ ] one page should occupy the full width of the (mobile) screen.
- [ ] the navbar should be responsive like in [this example](https://codepen.io/hanlinC/pen/dJwil)
- [ ] buttons should be bigger (suitable for any fingers)
- [ ] switch font-size in `em` and `rem` . Use `px` only for html/body tags
## UX mobile gestures specs

The [ mobile gestures](http://static.lukew.com/TouchGestureGuide.pdf) should be defined properly following this grid (those are examples of UX specs, open to discussion of course) 
- `tap`: does nothing
- `double tap`: start editing 
- `pan up/down` : scroll (default)
- `pan left/right` : scroll vertical ?
- `swipe left/right` show next/prev page
- `press` after a long press, the user can drag & drop paragraph.  it does not prevent user from simply scrolling
- `pinch / spread` : increase / decrease size (but stay on a single page)
## Events handling

This is quite tricky as there is already several ways to handle events, leading to possible conflicts in propagation etc. One of the problems with wiki-client current code base is that is has events handling in different parts of the code. 

There is different ways of solving this, one being to move all events into a separate `events.coffee` file. This can rapidly become painful to maintain. Another way, which I personally like more, will be to clearly separate UI components (it is already done quite well) and chose a single component structure or way to handle them all. Moving to a recent framework like React will save a lot of hassle but of course it is a lot of work and better suited for a 2.0 version - if we are eager to work on this.


opn:
I've made a wiki page to explore these UX issues and added some suggestions here:
* http://future.fedwiki.org/view/mobile-gestures-for-wiki

PR #174, opened 2016-09-27
updating video drop handler, and adding tests

paul90:
A reworking of the drop video handling.

Rather than performing a series of regular expressions this code uses node's `url.parse` to separate the part of the URL and switch on the `hostname` from the dropped URL.

Tests are also added.


WardCunningham:
Thanks for the tests.

paul90:
Published as part of + wiki-client@0.8.1

PR #176, opened 2016-11-17
use protocol-relative links when making requests - first pass

paul90:
The start of making the client work over https

There are updates to a number of plugins that are part of this update.

### DIFF
--- a/lib/future.coffee
+++ b/lib/future.coffee
@@ -13,7 +13,7 @@ emit = ($item, item) ->
   if transport = item.create?.source?.transport
     $item.append """<br><button class="transport" data-slug=#{item.slug}>create</button> transport from #{transport}"""
     $item.append "<p class=caption> unavailable</p>"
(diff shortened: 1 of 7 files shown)


lib/future.coffee
```diff
@@ -13,7 +13,7 @@ emit = ($item, item) ->
   if transport = item.create?.source?.transport
     $item.append """<br><button class="transport" data-slug=#{item.slug}>create</button> transport from #{transport}"""
     $item.append "<p class=caption> unavailable</p>"
-    $.get 'http://localhost:4020', ->
+    $.get '//localhost:4020', ->
```
WardCunningham: https://github.com/fedwiki/wiki-client/pull/155 that introduced this code includes a picture.

paul90: It certainly look very strange - a hard coded port number is not good.

paul90: There is also the related issue of listening for http and https requests on the same port.

PR #178, opened 2016-12-26
padding to stop jitter on button hover

WardCunningham:
Recent chrome browsers will inappropriately add a scrollbar to the journal upon hover over the the add or fork buttons when they are alone on the last line. (Firefox does not exhibit this unwanted behavior.)

image

This case is easy to duplicate by retrieving from history a revision where this is true. 

The fix is to restore some padding that was removed with the introduction of the .paper div. This leaves room for the hover effects which must be of box dimensions that round differently between chrome and firefox.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -152,7 +152,7 @@ footer {
   clear: both;
   background-color: #eeeeee;
   overflow: auto;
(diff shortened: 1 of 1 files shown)


paul90:
+ wiki-client@0.8.3

PR #180, opened 2017-01-13
ignore links with empty slugs

WardCunningham:
We depend on slugs to give us case insensitivity. That is [[Welcome Visitors]] and [[welcome visitors]] link to the same page. We also get a sometimes unwanted insensitivity to all non-alphanumeric characters.

This pull request refuses to link to a title with no alphanumeric characters. This avoids the case where there the link slug would be empty. We don't yet consider the case where the slug is non-empty but stripped of useful identifying unicode characters.

image

This improvement based on conversation on matrix:
https://riot.im/app/#/room/#fedwiki:matrix.org/$14842628142654GcOwU:matrix.allmende.io

### DIFF
--- a/lib/resolve.coffee
+++ b/lib/resolve.coffee
@@ -38,7 +38,10 @@ resolve.resolveLinks = (string, sanitize=escape) ->
 
   internal = (match, name) ->
     slug = asSlug name
-    stash """<a class="internal" href="/#{slug}.html" data-page-name="#{slug}" title="#{resolve.resolutionContext.join(' => ')}">#{escape name}</a>"""
+    if slug.length
+      stash """<a class="internal" href="/#{slug}.html" data-page-name="#{slug}" title="#{resolve.resolutionContext.join(' => ')}">#{escape name}</a>"""
+    else
+      match
 
   external = (match, href, protocol, rest) ->
     stash """<a class="external" target="_blank" href="#{href}" title="#{href}" rel="nofollow">#{escape rest} <img src="/images/external-link-ltr-icon.png"></a>"""


paul90:
from matrix:

> This refusing to link if the sanitized link is empty is I think a first step. Filenames is node.js are UTF-8 strings, so using a sanitized UTF-8 string as the filename may be a good next step (though the string needs converting to lowercase) - there is a filename sanitizer https://github.com/parshap/node-sanitize-filename<br/>This probably also means the URL containing a URL encoded version of the page title, rather than the slug.

Migrating to a different filename generation has an interesting problem with being able to support different schemes in the client - as not all servers will upgrade at the same time, and some may never.

This will require some careful thought...

WardCunningham:
I want to be sure that we don't establish a precedent with this commit that we regret when we do handle non-alphanumeric and unicode titles more completely. I had a brief flash of doubt but it has passed. I will explain.

My concern was that we might one day cause existing text to fall under a new interpretation. For example, if I wrote [[$%^&]] because I wanted exactly that and found that I got exactly that I would be satisfied. But if my text later morphed into $%^& and hyperlinked I might feel that I had been misled.

I will push on this idea a little further because it might lead to an important principle.

We could say that [[$%^&]] is to always be interpreted as markup and should print as $%^& even when rendered in our current environment where we can't access pages with that title. This would distinguish between the recognition of links and the retrieval of linked pages.

I now admit that this is to worry about small details when there are much more important decisions to be made. A more useful principle might be that of best effort: the client takes the information it has available, title or slug or both, and forwards it to a server hoping to get a page in return. If the server can find a likely page given what might be ambiguous information, then we have something to display and might signal alternatives as we now signal twins.

paul90:
Thinking about `[[$%^&]]` -- this raises the question of what to do with links, in general, that either using the current slug, or any future, sanitization yield an empty string. Currently `[[...]]`, and `[...]`, is always interpreted as a link - maybe we need a way to escape `[` `]` characters. While, initially this would allow links such as `[[$%^&]]` that give an empty slug have their text automatically altered to escape the square brackets. This will also allow the same text to continue to be written, though the square brackets would have to be manually escaped unless the code continued to look for empty *old style* slugs.

These *empty* slugs are only one case, and probably not the most problematic.

Currently, the slug is created from a page's title on the client and used as the name of the file we request from the server. Using a different approach will mean that a client probably should not generate the slug itself, but rather look it up in the sitemap file. The current client does not do this, so it is possible that wiki that have not been upgraded will not be able to see that content where the two schemes give a different slug.

Currently looking for twins is based on looking for pages in the neighbourhood that have the same slug. In a neighbourhood containing a mix of old and new servers pages that currently are considered twins need not have the same slug, so there needs to be some other mechanism for discovering twins. Maybe looking for some matching in the page title, probably ignoring case and punctuation. Or maybe we look for some common ancestry, which might mean a page with the same title might have a looser association than *twin* if there was no common ancestry.

Still lots to think about...

WardCunningham:
I will merge this small change which falls well short of what we could do but should not be discarded as we turn our attention to more pressing issues for the moment. I don't feel that this commit deserves a new release. Let it become part of a future release with more substance.

PR #181, opened 2017-01-30
site adapter prototype

WardCunningham:
This pull request introduces a family of "site adapters" which can compose and perform server facing operations. The prototype supports url-formatting and json-get from browser local storage and by the http protocol. The intention is that these can grow to support more network types and communication protocols, most urgently, https.

The wiki global provides three factories for creating lightweight adaptors.
- wiki.local
- wiki.origin
- wiki.site('fed.wiki.org')

Each adaptor supports (or will support) get, put and url formatting.
- url = wiki.origin.url 'favicon.png'
- wiki.origin.get 'welcome-visitors.json', (err, page) -> ...
- wiki.origin.put 'welcome-visitor/action', action, (err, reply) -> ...

The prototype temporarily adds `?adapted` to urls so that it is clear that requests observed in the browser have been issued through the adaptor mechanism.

image

We anticipate each factory will be implemented in its own module and required into the wiki.coffee module or bootstrapped by some yet to be determined method.

The page handler get logic demonstrates how adapters can be composed for more complex behavior such as polling different adaptors for the presence of a desired page. See the `localBeforeOrigin` composite adapter.

We hope that this mechanism easily integrates with the @paul90's http/https adaptor work already in progress. It is my intention to hunt down remaining server facing ajax requests once we have established that this is a suitable direction for our combined work.

### DIFF
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -22,8 +22,9 @@ module.exports = ($journal, action) ->
   else
     $action.appendTo($journal)
   if action.type == 'fork' and action.site?
+    url = wiki.site(action.site).url('favicon.png')
     $action
-      .css("background-image", "url(//#{action.site}/favicon.png)")
+      .css("background-image", "url(#{url})")
       .attr("href", "//#{action.site}/#{$page.attr('id')}.html")
       .attr("target", "#{action.site}")
       .data("site", action.site)
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -22,10 +22,11 @@ module.exports = ($journal, action) ->
   else
     $action.appendTo($journal)
   if action.type == 'fork' and action.site?
-    url = wiki.site(action.site).url('favicon.png')
-    $action
-      .css("background-image", "url(#{url})")
-      .attr("href", "//#{action.site}/#{$page.attr('id')}.html")
-      .attr("target", "#{action.site}")
-      .data("site", action.site)
-      .data("slug", $page.attr('id'))
+    wiki.site(action.site).getURL 'favicon.png', (backgroundURL) ->
+      wiki.site(action.site).getURL "#{$page.attr('id')}.html", (forkedPage) ->
+        $action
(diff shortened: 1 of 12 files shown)


WardCunningham:
I've searched wiki-client for remaining site references. I've added a commit that handles with site adaptors two out of three gets. The remaining get is related to the Transport plugin which needs work of its own.

Puts will require an equivalent amount of work in pageHandler which I will address soon enough. We don't put to remote sites so it will only require support in wiki.local and wiki.origin. 

As I understand how we plan on proxying https=>http we will want to detect that need in the adaptor factory method `wiki.site(domain)` which should be the only factory that needs this logic.

WardCunningham:
@paul90 I'd like to confirm that we are on the right track before I go through all of the plugins looking for gets and puts. Let me know if this integrates well with the code you showed me last weekend. Thanks.

paul90:
Not sure - just managed to loose the my work in GIT while trying to merge in the latest update :(

The process of determining the correct url is async, so `url` will need a callback - at which point the code calling it gets very messy. Rather than adding the callback to `url` I added `getURL`.

paul90:
I from memory I managed to break the neighborhood flags, they were working but then...

WardCunningham:
Ah, the neighborhood flags ...

That is probably some of our most intricately async code in the client. Working async code can be magical but broken async code is surely mystifying. Let me work through it the getURL call. I will channel Pete Hodgson as I do. If I have to back out changes I will make my own async getURL that uses a timer to test my own async logic.

paul90:
Almost looks as if some of the callbacks are going astray - I've got it adding some of the flags in the neighborhood bar. But, one the first in any group is getting added - e.g. when adding a number of sites using a roster...

paul90:
I notice that a number of the files had `wiki = require './wiki'` missing

currently it seems to be complaining that `wiki.local` is undefined, even though it is defined...

paul90:
in #182 I'm splitting the bits out to try and isolate things that are not working, and try and fix them

paul90:
closed in favour of #185

PR #182, opened 2017-02-08
site adapter prototype - take 2

paul90:
I'm finding it difficult to see what is going on in #181, so this is an other look at the same approach. But only changing one thing at a time, and I have added a lot more diagnostic `console.log` to try and pinpoint what is going on. At least part of the problem is that there are functions within functions and the results don't appear to being passed out.

### DIFF
--- a/lib/neighborhood.coffee
+++ b/lib/neighborhood.coffee
@@ -3,6 +3,7 @@
 # slowly and keeps track of get requests in flight.
 
 _ = require 'underscore'
+siteAdapter = require './siteAdapter'
 
 module.exports = neighborhood = {}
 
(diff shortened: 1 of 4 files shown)


paul90:
Current issue, if more that one site is being added at a time, via a roster but probably also by other mechanisms - only the first of those site's flags is getting displayed in the neighborhood bar.

paul90:
Filtered console.log shows...

```
about to trigger new-neighbor localtest.me
in new-neighbor, about to call getURL localtest.me  
in new-neighbor /favicon.png?adapted  
about to trigger new-neighbor-done localtest.me  
about to trigger new-neighbor test.localtest.me  
in new-neighbor, about to call getURL test.localtest.me  
about to trigger new-neighbor test1.localtest.me  
in new-neighbor, about to call getURL test1.localtest.me  
about to trigger new-neighbor test2.localtest.me  
in new-neighbor, about to call getURL test2.localtest.me  
about to trigger new-neighbor test3.localtest.me  
in new-neighbor, about to call getURL test3.localtest.me  
about to trigger new-neighbor test4.localtest.me  
in new-neighbor, about to call getURL test4.localtest.me  
in new-neighbor //test.localtest.me/favicon.png?adapted  
about to trigger new-neighbor-done test.localtest.me  
about to trigger new-neighbor-done test.localtest.me 
about to trigger new-neighbor-done test1.localtest.me
about to trigger new-neighbor-done test2.localtest.me
about to trigger new-neighbor-done test3.localtest.me
about to trigger new-neighbor-done test4.localtest.me
```

looks as if we are only getting into the callback from `getURL` in `new-neighbor` once

paul90:
that seems to have sorted the problem I had seen, back to #181

PR #184, opened 2017-02-13
site adapter with a spot to patch wrong flags

WardCunningham:
With this version I try to manage the case analysis and leave room for patching up flags that are not properly generated with the sync url function.

Missing code is marked with comments.

I've added an adapter that can serve pages out of an export file. I haven't tested this yet because I had trouble setting cors headers on readily available servers. Damn apache config.


paul90:
closed in favour of #185

PR #185, opened 2017-02-14
site adapter prototype - just flag and get

paul90:
This approach is one that @WardCunningham and myself discussed over on Sunday.

The problems with the previous prototypes start when we request a flag, and don't yet know how to construct the request for it. Determining how to correctly connect to a remote site is asynchronous, which forces get the flag's URL into being asynchronous, which then ripples back through the client.

Here, rather than determine the form of the URL for the flag, if we don't already know how to construct its URL we create a temporary flag, which we will use for the remote site until know to construct the request for the remote site. Once we know the URL for a remote site's flag, we will replace its temporary flag with the real one.

screen shot 2017-02-14 at 13 24 37

The first commit here will just use a temporary grayscale gradient flag for a remote site. It does not yet have the logic for determining the correct route to a remote site, replacing the temporary flags, or fetching remote resources.

### DIFF
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -23,7 +23,7 @@ module.exports = ($journal, action) ->
     $action.appendTo($journal)
   if action.type == 'fork' and action.site?
     $action
-      .css("background-image", "url(//#{action.site}/favicon.png)")
+      .css("background-image", "url(#{wiki.site(action.site).flag()}")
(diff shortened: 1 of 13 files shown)


paul90:
This will now as part of the flag request initiate finding the correct prefix to use for a site, and having found it replace any temporary flags used with the correct ones.

paul90:
Have added fetching page and sitemap using the adapter. This appears to be working very nicely.

PR #187, opened 2017-04-02
toggle wiki editing

WardCunningham:
This pull request is a variation of wiki@next that makes wiki editing optional.

Here with `wiki` not selected. Drag just selects text. No journal or action buttons.
image
Click `wiki` in the footer to restore normal wiki edit operations.
image

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -27,6 +27,7 @@ escape = (string) ->
 
 textEditor = ($item, item, option={}) ->
   console.log 'textEditor', item.id, option
+  return unless $('.editEnable').is(':visible')
 
   keydownHandler = (e) ->
 
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -219,7 +219,7 @@ $ ->
     .click ->
       $('.editEnable').toggle()
       $('.page').each refresh.cycle
-  $('.editEnable').toggle()
+  $('.editEnable').toggle() unless isAuthenticated
 
   target.bind()
 
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -212,6 +212,15 @@ $ ->
     .click ->
       dialog.open "Lineup Activity", lineupActivity.show()
 
+  # $('.editEnable').is(':visible')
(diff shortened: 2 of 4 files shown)


WardCunningham:
This feature is not tied to the user's ownership or login status. We talked about that being the only flag. I took the second flag approach to avoid coding and usability complications.

We might alternatively consider:
- default to wiki enabled if logged in.
- use touch-compatible tap sequences to enable edit.
- make always-on and always-off administrative options.

WardCunningham:
The refresh that this invokes is unexpected and drives the dom and lineup further out of sync. This is noticeable in that ghost future pages are variable in their titling. The inspector console shows that the dom and lineup don't even agree as to the number of pages.

WardCunningham:
This feature is available for testing on a server running wiki@next. In my configuration I installed this with these commands (from memory):
```
npm install -g wiki@next
cd /usr/local/lib/node_modules/wiki
npm install wiki-client@optional
```

opn:
Good experience so far working with this early version. Reading on mobile almost works perfectly. Here are some suggestions / feedback:

1. The `wiki` toggle would ideally be integrated into the `padlock` interaction
2. Logging in should enable automatic `wiki` editing
3. While we have both `padlock` and `wiki` toggles, and if we enable automatic `wiki` editing on login, then we should rename the `wiki` toggle `annotate` and only show it on origin sites we are not logged in to.

Another thing we could experiment with is adding left and right swipes for (mobile) touchscreen navigation of the lineup. While navigating through clicking on links now works well on mobile, you cannot easily go back or forwards in the lineup. Swipe is the gesture we need for that.

WardCunningham:
I would be against integrating the lock behavior with the edit behavior since they operate by very different means in different layers of the application. The lock is "elevated" to the footer by intricate client-server protocol while the wiki control is "lowered" to the footer from page rendering by one-time startup code. Residence in the footer is all they have in common.

I have no plans to add further mechanism to this mode switch. I would be much more interested in perfecting modeless gestures that work well in all environments but will leave that for subsequent pull requests. I would be happy to have this merged as is.

paul90:
All right, lets merge this but I look forward to a modeless version.

PR #191, opened 2017-04-23
rebuild existing pages on toggle wiki

WardCunningham:
In an earlier comment I noted that toggling read-write would cause an unwanted accumulation in the lineup model. https://github.com/fedwiki/wiki-client/pull/187#issuecomment-291378581

I've tracked this problem down to a call into lib/refresh at too high a level. The lineup is intact so we should instead just rebuild the dom $page from available information. There is an entry point for exactly this purpose and a sample call a little higher in the code where the page `create` button rebuilds an existing (ghost) page.
```
.delegate 'button.create', 'click', (e) -> ...
```

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -218,7 +218,10 @@ $ ->
     .appendTo('footer')
     .click ->
       $('.editEnable').toggle()
-      $('.page').each refresh.cycle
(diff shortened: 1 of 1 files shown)


paul90:
published with 'next' tag as `0.9.0-4`

PR #192, opened 2017-04-30
simplify footer

WardCunningham:
We label the search box inside the box using the input field's placeholder attribute.
(Some experts suggest a label, box, magnifier icon and button. Yuck.)

screen shot 2017-04-29 at 9 09 07 pm

We also discard the ≡ hamburger menu which never developed into a menu. The one function it offered was a crude timeline visualization of journal edits which never proved useful for anything.


paul90:
Vaguely remember working on something to add to a ≡ menu - I wonder what it was....

PR #193, opened 2017-04-30
simplifying the footer source code

WardCunningham:
Darn. I shouldn't commit late at night. This morning I found an editor window open and not saved. Please accept this late bit of cleanup.

Here I delete the code that I had previously just commented out. Git will keep it for us if we ever want to revive the hamburger menu.


WardCunningham:
Oops. Did that wrong. Shouldn't commit in the morning it seems. I've been working against next. I guess this will merge ok but it isn't easy to find my one commit.

WardCunningham:
I think it best if I retract this and pull request again.

PR #197, opened 2017-07-04
detect and process page delete

WardCunningham:
Work in progress. See checklist below.

After six years, finally working on delete command. To delete a page, recall its creation with no content, then fork that to the server. New logic interprets forking a page with no contents as delete. (This is like saving a Paragraph with no text to delete it.)

kwecxeerxsqunafybrvqhsjr

After a successful server-side DELETE, the page appears as it did before it was created. The create action in the future journal is new but preserves some parameters from the old such as the Transporter continuation logic, if any.

pgrgbhkmbbpvsgmzweaqcpeb

So far we have discovered and addressed many client-side issues while simulating server side delete in pageHandler.delete with a timer and fixed value for err status. We either delete the file with a shell command or just reload to discover that the deleted page has reappeared.
```
(cd ~/.wiki; rm pages/testing-fork-to-delete; rm status/sitemap*)
```

More to do client-side:
- [ ] handle delete-neighbor-done events to adjust twins and recent changes
- [ ] handle cases where page is in browser local storage
- [ ] handle non-origin and not-logged-in cases
- [ ] test with downstream dependents within lineup
- [ ] turn other lineup copies to ghosts, check that they can be forked to undelete
- [ ] consider longer-term undelete scenarios from server-side recycling
- [ ] test with Transporter that uses create continuations
- [ ] improve Future plugin's wording for the just-deleted case
- [ ] delete recycled page

We will expect a coordinated update to wiki-server.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -290,7 +290,7 @@ p.readout {
 }
 
 .page.recycler > .paper {
-  box-shadow: inset 0px 0px 40px 0px rgba(107,142,35,.5)
+  box-shadow: inset 0px 0px 40px 0px rgba(220,0,0,.5)
 }
 
 .paper {
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -289,6 +289,10 @@ p.readout {
   box-shadow: inset 0px 0px 40px 0px rgba(0,180,220,.5);
 }
 
+.page.recycler > .paper {
+  box-shadow: inset 0px 0px 40px 0px rgba(107,142,35,.5)
+}
+
 .paper {
   /*width: 430px;*/
   padding: 30px;
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -139,16 +139,29 @@ $ ->
       $page = $(e.target).parents('.page')
       return if $page.find('.future').length
       pageObject = lineup.atKey $page.data('key')
-      action = {type: 'fork'}
-      if $page.hasClass('local')
-        return if pageHandler.useLocalStorage()
-        $page.removeClass('local')
-      else if pageObject.isRemote()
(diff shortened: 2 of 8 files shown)


WardCunningham:
Why have we not offered delete already?

It seems obvious that if you can add pages, you should be able to delete them. This is most certainly true when adding pages to a traditional document. But hypertext expands to meet all needs so one just moves on. Federation relieves the pressure to get it right even more. No need to make one site perfect. Just make another and fork the best content forward. We've learned from living with this limitation.

We added delete for browser local pages early on. Here an accidental fork would hide real content behind yellow halo pages. This case is much easier because we knew there would not be dangling references. Dangling references and undelete support remain important issues to be explored for regular pages now.

Why is now the right time to work on delete?

We now have a client-side abstraction for different kinds of pages called the siteAdapter. Up until now we have special-cased every page type: normal, remote, local, and plugin pages require if-statements throughout the core client codebase. We're hoping that siteAdapter will allow a collection of deleted pages to slide into our system like any other site offering content to an author. We'll see.

There is also the sheer complexity of removing things that get used elsewhere. Our first commit here touched four separate and individually complex modules. This was not the sort of thing one would want going on while tackling even more important revisions to the core. With those, like pluggable security and https proxy to the federation out of the way, delete becomes a top priority.

See [Top Maintenance Priorities](http://ward.asia.wiki.org/top-maintenance-priorities.html)

So now we are ready to give delete the serious attention it deserves.

paul90:
As mentioned in the hangout, some initial work on the recycler are available. See, 07f935c

paul90:
Merging in the client updates for the recycle

When an empty page is forked from recycle it is deleted.

Issue #199, opened 2017-10-02
Fork from transporter incorrectly seen as delete


WardCunningham:
We suspect complains in the chat can be traced to incomplete logic in #197.

Regarding the trouble forking transported pages, If the screen shot provided is associated with that failure, I notice that DELETE that failed is probably not what you were trying to do. Perhaps when you were trying to fork the transported ghost page into your site, new logic though you were trying to delete the (not yet existing) page. [riot](https://riot.im/app/#/room/#fedwiki:matrix.org/$15069574022109395RuxVa:matrix.org)

At the least, a fork of a ghost page without rev addressing should not be interpreted as a delete no matter what condition the journal is in (transporters can be lazy).

opn:
I've made some progress tracking this issue down. For test purposes you can try the examples provided on the following [page](http://issues.fedwiki.org/view/welcome-visitors/view/broken-transporters)

The issue at least for the OneTab transporter is that the newly created page has a minimal journal entry rather than a full history of the creation of every element. I assume this is triggering a "delete page" action as per the new logic.

The MediaWiki Transporter also has an creation only journal - so this makes sense. I could get around this issue by restoring the full journal history creation in the transporter. However this might be a good time to consider issues and benefits of Journal simplification?

paul90:
This was closed by #200

PR #200, opened 2017-10-03
detect delete based on _rev0

WardCunningham:
Fix for issue #199.

Look for _rev0 which shows that user really did click an action, not just that the ghost page has journal with only create.


paul90:
Looks good - best to be defensive against transporters that don't create a journal

PR #201, opened 2017-10-03
More concise synopsis.

paul90:
We take the first paragraph of the candidate synopsis, and then reduce it to the first 560 characters.

This is a proposed fix for fedwiki/wiki-server#136

### DIFF
--- a/lib/synopsis.coffee
+++ b/lib/synopsis.coffee
@@ -14,4 +14,6 @@ module.exports = (page) ->
     synopsis ||= page.story? && "A page with #{page.story.length} items."
   else
     synopsis = 'A page with no story.'
-  return synopsis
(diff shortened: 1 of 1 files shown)


WardCunningham:
Ah, nice thought on the line break. They only show up when pasting bulk text so this fix directly address the problem of huge pastes. 

I do sometime get strange newlines included in text copied out of pdf documents so I will have to watch for that. I wonder if there is a css option that will make newline visible within a textarea?

paul90:
> I do sometime get strange newlines included in text copied out of pdf documents so I will have to watch for that. I wonder if there is a css option that will make newline visible within a textarea?

Maybe using `white-space: pre-wrap;` on the textarea would help...

Issue #202, opened 2017-10-06
Bug shift creating from an internal link


opn:
When trying to create a new page by shift-clicking on an internal link for a page that does not exist the page title is named after the domain name of the site of origin instead of the page title in the Future Plugin:

<img width="996" alt="screenshot 2017-10-06 17 54 41" src="https://user-images.githubusercontent.com/80854/31289084-83ba62b6-aabf-11e7-8c4f-8c9abd080244.png">

paul90:
Fixed in wiki-client@0.10.3

PR #203, opened 2017-10-07
be specific looking for internal link text

WardCunningham:
Fix for issue #202 

When pageHandler can't find a page the refresh logic builds a Future as substitute. This logic looks for the internal link from which to recover the proper title. This commit revises the jquery selector used to be specific that it wants links of class .internal  so that any intervening pages with the same link in its flag does not confuse the logic.

image

Thanks go to @opn for recognizing the repeatable case.

### DIFF
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -335,8 +335,9 @@ cycle = ->
   }
 
   whenNotGotten = ->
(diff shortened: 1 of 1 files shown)


paul90:
published as + wiki-client@0.10.3

PR #204, opened 2017-10-25
split markdown items when enter is pressed twice

paul90:
This is part of a re-look at the markdown plugin. Here we start to treat markdown items a bit more like paragraph items.

When editing a markdown item:

* pressing `enter` twice will save the current markdown item and create a new one. (this will also split a markdown item is the insertion point is not at the end of the item).
* pressing `backspace` at the start of a markdown item will append the current markdown item onto the previous page item, if it is also a markdown item.

**N.B.** this does not prevent a markdown item from containing more than block of markdown text.

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -27,6 +27,7 @@ escape = (string) ->
 
 textEditor = ($item, item, option={}) ->
   console.log 'textEditor', item.id, option
+  enterCount = 0 if item.type is 'markdown'
   return unless $('.editEnable').is(':visible')
 
   keydownHandler = (e) ->
@@ -43,13 +44,13 @@ textEditor = ($item, item, option={}) ->
       return false
 
     # provides automatic new paragraphs on enter and concatenation on backspace
-    if item.type is 'paragraph'
+    if item.type is 'paragraph' or 'markdown'
       sel = getSelectionPos($textarea) # position of caret or selected text coords
 
       if e.which is $.ui.keyCode.BACKSPACE and sel.start is 0 and sel.start is sel.end
         $previous = $item.prev()
         previous = itemz.getItem $previous
-        return false unless previous.type is 'paragraph'
+        return false unless previous.type is item.type
         caret = previous[option.field||'text'].length
         suffix = $textarea.val()
         $textarea.val('') # Need current text area to be empty. Item then gets deleted.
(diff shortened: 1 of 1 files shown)


lib/editor.coffee
```diff
@@ -43,34 +44,41 @@ textEditor = ($item, item, option={}) ->
...
+        # console.log "Type: #{item.type}, enterCount: #{enterCount}"
         return false unless sel
-        $page = $item.parents('.page')
-        text = $textarea.val()
-        prefix = text.substring 0, sel.start
-        suffix = text.substring(sel.end)
-        if prefix is ''
-          $textarea.val(suffix)
-          $textarea.focusout()
-          spawnEditor($page, $item.prev(), prefix, true)
```
paul90: The `true` in this call looks to be redundant, `spawnEditor` only has 3 parameters, and contains no code to retrieve any extra arguments, so it has been removed.

WardCunningham:
I have been wondering how to parameterize the text editor. This seems a more immediately useful approach instead.

paul90:
Published as + wiki-client@0.11.0

WardCunningham:
Some observations comparing Paragraph to Markdown split/join behavior.

With carrot cursor at end of text:
- Paragraph: a join cancels a split with no actions sent and no journal modification.
- Markdown: a join does not cancel a split because one newline has already been recorded.

With carrot cursor at beginning of text:
- Paragraph: a split opens an empty editor in front of the unmodified item.
- Markdown: a split adds (prepends) an empty item and reopens the editor on the unmodified item.
(the empty item contains one newline but cannot be seen, opened or removed)

With carrot cursor in middle of text:
- Paragraph: a split saves the prefix as edit action and opens unsaved editor with suffix.
- Markdown: a split saves the prefix with newline as edit action and opens unsaved editor with suffix.

With last word selected:
- Paragraph: a split saves the prefix and opens empty unsaved editor.
- Markdown: a split saves the prefix with newline and opens empty unsaved editor.

With first word selected:
- Paragraph: a split saves the suffix and opens empty unsaved editor in front of it.
- Markdown: a split adds (prepends) an empty item and reopens the editor on the item suffix.

With middle word selected:
- Paragraph: a split saves prefix, adds suffix, and opens empty unsaved editor between them.
- Previously: a split saves the prefix and opens unsaved editor on suffix.
- Markdown: a split saves the prefix with newline and opens unsaved editor on sufix.

I have tested this by installing wiki: 0.12.2, creating a test page with similar text as Paragraph and Markdown, and then forking from history back to this state before each test. I use the `json` link to observe the modified story and journal. I use shift-hover over actions to confirm that item identities are sensibly preserved.

I believe that many of these differences can be resolved by removing the newline added with the first enter once the second enter has been detected. This should avoid unwanted updates to thus unchanged items.

I am alarmed that the "split middle word paragraph" case performs two simultaneous actions. This is a regression from the previous version of the editor and risks client-server synchronization problems on slow connections. Could this be related to https://github.com/fedwiki/wiki-client/pull/204#discussion_r146853959?

I apologize for not conducting this testing earlier. This was simply my laziness installing branches.

WardCunningham:
I previously noted a regression for splitting paragraphs in the "middle word selected" case. I now believe that to be a false alarm probably due operator fatigue running through all these cases. I've carefully tested both old and new code under development conditions and find no differences for type Paragraph. I've also produced a clean, side-by-side, diff and find no hint of a change to the flows for type Paragraph. CoffeeScript's semantic whitespace had interacted with git's diff to throw me off. Here is the diff I found more useful:
```
diff -by -W 180 lib/old-editor.coffee lib/editor.coffee
```

WardCunningham:
With my continued reading I did find one suspicious statement. I examined the generated code in the chrome inspector. If it is to be trusted, I believe this if statement will always pass, which is to be expected of the two test cases I have chosen.

if item.type is 'paragraph' or 'markdown'
```
if (t.type, !0) {
```

I believe this is the intended statement. I will make this change and offer a pull request.

if item.type is 'paragraph' or item.type is 'markdown'

```
if ("paragraph" === t.type || "markdown" === t.type) {
```

PR #205, opened 2017-11-12
proper form for alternative

WardCunningham:
Discussed in https://github.com/fedwiki/wiki-client/pull/204#issuecomment-343762247


paul90:
Published as + wiki-client@0.11.1

PR #206, opened 2017-11-12
simple alt-m handler, might be best

WardCunningham:
Type cmd/alt-m to return to plugin menu.

- recognized during edit only
- menu does not erase text, returns to text editor on most selections
- further text change required to persist choice

Typical workflow:
- recognize preferred plugin while editing
- cmd/alt-m to select desired plugin
- continue editing then cmd/alt-s to save

Preferred mechanism for adding captions to images:
- split into the editor
- write the desired caption
- cmd/alt-m to return to menu
- drop image on the menu

image
image
image

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -32,15 +32,21 @@ textEditor = ($item, item, option={}) ->
 
   keydownHandler = (e) ->
 
-    if (e.altKey || e.ctlKey || e.metaKey) and e.which == 83 #alt-s
+    if (e.altKey || e.ctlKey || e.metaKey) and e.which == 83 #alt-s for save
       e.preventDefault()
       $textarea.focusout()
       return false
 
-    if (e.altKey || e.ctlKey || e.metaKey) and e.which == 73 #alt-i
+    if (e.altKey || e.ctlKey || e.metaKey) and e.which == 73 #alt-i for information
       e.preventDefault()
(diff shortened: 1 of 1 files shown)


opn:
With double enter to split markdown I have a question. One value of using markdown is the ability to easily have long paragraphs for a section that you may wish to drag and drop or delete at some later stage. These "long markdown sections" are useful and I may wish to type them.

Also when I make markdown menus ( a series of bulleted linked markdown items) I often will insert a blank line by inserting a space on it's own line.

In both these cases will I no longer be able to author long a markdown sections on my laptop without an enter key?

paul90:
For markdown plugin items the split happens when you press `enter` twice, consequetively.

> A paragraph is simply one or more consecutive lines of text, separated by one or more blank lines. (A blank line is any line that looks like a blank line — a line containing nothing but spaces or tabs is considered blank.) Normal paragraphs should not be indented with spaces or tabs.

A new text block can be created, within a markdown plugin item, by pressing `enter`, `space`, `enter`. This does not cause the markdown plugin item to be split, but will separate blocks of text with it.

WardCunningham:
@opn This pull request is unrelated to the Markdown plugin other than by the coincidence that Markdown is one of the choices on the menu it provides.

paul90:
Published as + wiki-client@0.11.1

PR #207, opened 2017-12-01
Change the site adapter discovery mechanism

paul90:
Here we switch to using a queue to prevent overloading the browser, or a remote server, and using ajax to GET the image which allows us to detect any redirects that will cause problems later in the flow.

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -118,9 +118,15 @@ $ ->
       finishClick e, name
 
     .delegate 'img.remote', 'click', (e) ->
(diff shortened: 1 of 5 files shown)


paul90:
Something to be aware of, this new code has problems if a foreign server does not set CORS headers on the response when fetching the favicon.

paul90:
In 45e038f clicking on the flag of a neighbor that fails to load its sitemap will refresh the route to that site.

This probably should also cause a retry at loading the site's sitemap...

paul90:
Published as + wiki-client@0.12.0

PR #208, opened 2017-12-10
wide margins for print

WardCunningham:
Sometimes our short paragraphs and narrow column spill ungracefully onto the printed page. Better to add broad margins and retain some approximation of the online look.

image

Becomes:

image

From the original:

image

### DIFF
--- a/client/style/print.css
+++ b/client/style/print.css
@@ -9,7 +9,7 @@ body {
 .main {
 	width: auto;
 	border: 0;
-	margin: 0 5%;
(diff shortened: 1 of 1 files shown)


WardCunningham:
Oops. Screen shots were in landscape mode. This modification looks even better in portrait mode.

image

paul90:
Screen shot with the restyling of links

screen shot 2017-12-11 at 09 46 28

paul90:
Included in + wiki-client@0.12.0

PR #210, opened 2017-12-10
avoid trailing space in flag anchor

WardCunningham:
This makes a difference in the appearance of tab-sequence highlights.

Without this change:
image

With this change:
image


paul90:
Included in + wiki-client@0.12.0

PR #215, opened 2018-02-18
add license details to page footer license link

WardCunningham:
@opn reminds us that the attribution from forks in the journal is hidden in read-only mode and no attempt has been made to explain sources entering by transporter.

Here we intercept the previously supplied link to Creative Commons and offer our interpretation of the journal in the context of our chosen license.

We look for non-fork actions and report the sites that they appear to originate at based on subsequent forks or the ultimate owner of the page.

image

We look for "provenance" attached to the first action (typically a Create) and render that as wiki text when present. Here I edit .wiki/localhost/pages/welcome-visitors to test this feature.

```
  "journal": [
    {
      "type": "create",
      "item": {
        "title": "Welcome Visitors",
        "story": []
      },
      "date": 1420938191608,
      "provenance": "This page provided by the authors of federated wiki. See [https://github.com/orgs/fedwiki/people github]"
    },
```

image

### More To Do
- [x] Make proper references to SiteAdapter rather than directly generating links.
- [x] Remove console.log.
- [x] Possibly add site flags to each site mention.

### Won't Do
- [ ] Possibly retrieve the author's name from the who field of welcome-visitors.

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -12,6 +12,7 @@ drop = require './drop'
 dialog = require './dialog'
 link = require './link'
 target = require './target'
+license = require './license'
 
 asSlug = require('./page').asSlug
 newPage = require('./page').newPage
@@ -101,6 +102,12 @@ $ ->
     return false
 
   $('.main')
+    .delegate '.show-page-license', 'click', (e) ->
+      e.preventDefault()
+      $page = $(this).parents('.page')
+      title = $page.find('h1').text().trim()
+      dialog.open "License for #{title}", license.info($page)
+
     .delegate '.show-page-source', 'click', (e) ->
       e.preventDefault()
       $page = $(this).parents('.page')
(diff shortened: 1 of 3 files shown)


paul90:
Have added using the site adapter for the links in f43953b 

Not sure that retrieving the author's name from the wiki page is going to be that simple. There are probably enough people that either replace that paragraph with something else, or put more than just their name in there.

WardCunningham:
Links are nice addition. Flags would be overkill and destroy the "document" look of the dialog. This pull request looks ready to go so I will removed the warning and surplus check boxes.

WardCunningham:
I'll fix the check-list. No reason to undue anything.

paul90:
Published as + wiki-client@0.14.2

opn:
Thanks for this - it's a great improvement.

Question - why not link directly to the wiki-page on the authors site rather than generally to the domain - a citation / Creative Commons reference should link to the original content - not just the author? Or is this planned for / too hard for now?

- https://wiki.creativecommons.org/wiki/Best_practices_for_attribution

Issue #217, opened 2018-07-16
Forking from recycler shows up in search


paul90:
Ward rasies an issue with recycler showing up in search on [matrix](https://matrix.to/#/!ORfrUEFeWFcHAMLFLr:matrix.org/$1531627108753224CozAx:matrix.org)

> Search encounters "recycler" as a domain name of a site and shows it as something new. The Roster plugin doesn't recognize it as a site so just prints it instead.
image.png
> ![image](https://user-images.githubusercontent.com/1552489/42744468-a10a622e-88c4-11e8-84f8-00faeb2c83eb.png)
> Somewhere we maintain a list of keywords that aren't sites, like view and local. These are client-side keywords but recycler seems to have made it to the server and it's sitemaps. I'm looking for the smoking gun now.
>  ![mkpvulligowebkiijxjwiepg](https://user-images.githubusercontent.com/1552489/42744482-bedd455a-88c4-11e8-9124-e74b5b3172b6.png)
> Fork from browser local storage doesn't identify a site. I'm not sure we gain anything by identifying a fork from the recycler. I'm surprised that I haven't seen this before. Do we want to keep it? I will have to hack my search engine if we do.

paul90:
We should be treating `recycler` the same as `view` and `local` when adding fork event to a page's journal.

https://github.com/fedwiki/wiki-client/blob/3e0a1dd499b5eb637eeb4a03970c9f0a4310e619/lib/pageHandler.coffee#L145

paul90:
Not yet managed to recreate the underlying issue...

Looking at the page journal for the page in question, something looks very strange, with the entries not all being in order. Notice that the create event is not the first entry.
![screen shot 2018-07-16 at 07 43 13](https://user-images.githubusercontent.com/1552489/42745788-7c252086-88cc-11e8-9620-452e53fd026a.png)
and in the page json
![screen shot 2018-07-16 at 07 49 14](https://user-images.githubusercontent.com/1552489/42745857-d3c744fe-88cc-11e8-8a50-9d5d8d18befe.png)

Clicking on the page history gives some very strange pages with `null` entries in the json. This looks to be caused by `move` entries referring to items that don't appear to exist due to the journal being out of order.

paul90:
Managed to recreate getting a `fork` journal entry with recycler as the site, by trying to edit the page in the recycler and then forking it back to the site.

[...]

PR #218, opened 2018-07-18
Splitout and perform any implicit fork

paul90:
Here we split out any implicit fork, and perform it, before applying the edit. We also correct the missing update to the current page state (removing the recycler halo). This addresses the case of a page in the `recycler` being edited, which previously with the compound action containing the implicit fork would try and apply to update to the current wiki page, rather than the page in the recycler.

`await` is used to sequence the two actions, this also saves rewriting `pushToServer` and `pushToLocal` to use callbacks and the associated ripple of changes and callback hell.

N.B. This does not address an associated issue with a *ghost* page being edited. Also not sure if this addresses all the potential problems raised in #217

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
   - "8"
   - "10"
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "6"
   - "8"
-  - "9"
+  - "10"
 install:
   - npm install
 before_install:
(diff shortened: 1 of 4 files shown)


paul90:
**N.B. To build this version of the client, Node 8 is required.**

WardCunningham:
The implicit fork and edit were bundled into one server turn around because this fixed some timing or synchronization bug from back in the ruby server days. Even with the node server we had timing troubles with hangout #1 participants in India. Some tweaks to the editor fixed these by again not doing server turn arounds in rapid succession. Let's discuss.

PR #219, opened 2018-09-16
add xhrFields for credentials

WardCunningham:
Companion to fedwiki/wiki-security-passportjs#18


paul90:
Published as + wiki-client@0.14.0

PR #220, opened 2018-09-19
modify siteAdapter to only send credentials when required

paul90:
* rename the existing key store to "routes"
* add new store "withCredentials" for if credentials are required. value only stored when true.

Within the remote site get we split out the ajax `GET` and recursively call it when we get a "Login Required" response and have not already got `withCredentials: true` set.

### DIFF
--- a/lib/siteAdapter.coffee
+++ b/lib/siteAdapter.coffee
@@ -6,8 +6,10 @@ localForage = require 'localforage'
 
 module.exports = siteAdapter = {}
 
-# we save the site prefix once we have determined it...
+# we save the site prefix once we have determined it,
 sitePrefix = {}
+# and if the CORS request requires credentials...
+credentialsNeeded = {}
 
 # when asked for a site's flag, if we don't know the current prefix we create
 # a temporary greyscale flag. We save them here, so we can replace them when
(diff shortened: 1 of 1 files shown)


WardCunningham:
I see this is working around my compromise of returning 200 instead of the expected 403.
```
if data.title is 'Login Required' ...
```
Perhaps a better approach is for me to send explanatory page json as a 403 and then modify the site adapter to display this when present. I only send 200 as a quick way around 403 not displaying anything. It might even be better to display as a ghost page since it isn't a real page. I'm unsure how much coding this would take but sending proper status codes would seem to be worth it.

WardCunningham:
Paul explained that the 200 is still required if an older wiki wanted to get the message.

paul90:
Yes, while using a `200` seems all wrong, and in many ways it is, the older client will handle the response and show the 'Login Required' content. Using a `403` would cause an older client react as if the page does not exist - but it would not give any indication that a login was required to see the content.

A couple of things that need some thoughts:

If the user is accessing the restricted page from a site with an older version of the client, login does not provide access as the client does not handle the cross-domain passing of credentials.

To the uninitiated there is no obvious indication of what they need to log into. Maybe needs a few extra words on the "Login Required" page to direct them to click on the page flag to get to the restricted site.

Think both those points can be covered by a few extra words on the "Login Required" or "Login to View" page. 

That aside, I think this PR is ready to be merged.

paul90:
Published as + wiki-client@0.14.1

PR #222, opened 2018-10-21
stop fall through

paul90:
`done` should only be called once, either in `getContent`, or in the else part of the condition.

This correct an error where done is getting called too soon, on the fall through, rather as part of the call back from getContent.


WardCunningham:
Thanks for spotting this. There is no way it can be anything but correct.

PR #226, opened 2019-02-20
dispatch and handle image urls

WardCunningham:
UNTESTED WORK IN PROGRESS

Considering https://github.com/fedwiki/wiki/issues/121

There really are times where you want a lot of large images on a page. We've talked about managing images in the assets directory but the logic is complex and remains unimplemented.

This pull request extends Factory to accept a dropped image url and creates an Image item with url set to that string. This makes a lightweight item that will not bloat a page even if the caption is repeatedly edited. Here we assume the image is persistently stored as one might expect from Wikipedia Commons or our own assets directories.

Workflow for adding small Image items:
1. Save image as local file with jpeg, jpg or png suffix.
2. Upload or drag image to Assets item, preferably with markup `page/<slug name of page>`.
3. Click image name to open it in new tab.
4. Drag that tab's location to a Factory to become an Image item.

Workflow for adding html item with img tag:
1. same
2. same
3. Copy Link Address of image name in Assets
4. Create an HTML item with markup like `>`

### DIFF
--- a/lib/drop.coffee
+++ b/lib/drop.coffee
@@ -24,6 +24,11 @@ isPage = (url) ->
     return item
   null
 
+isImage = (url) ->
+  if url.match /\.(jpeg|jpg|png)$/i
+    return url
+  null
+
 isVideo = (url) ->
   parsedURL = nurl.parse(url, true, true)
   # check if video dragged from search (Google)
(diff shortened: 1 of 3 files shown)


paul90:
Drag and drop of image urls has been implemented in #262. Closing this PR.

PR #228, opened 2019-03-22
Add hamburger menu to open system links

nrn:
paired with @WardCunningham 

Information here: http://ward.asia.wiki.org/view/hamburger-menu

image

image

### DIFF
--- a/lib/factory.coffee
+++ b/lib/factory.coffee
@@ -30,11 +30,12 @@ emit = ($item, item) ->
       <tr><td><ul id=format><td><ul id=data><td><ul id=other>
     """
     for info in window.catalog
-      column = info.category || 'other'
-      column = 'other' unless column in ['format', 'data']
-      menu.find('#'+column).append """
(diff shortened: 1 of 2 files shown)


nrn:
Forthcoming updates to activtiy and changes plugins should be a version bump such that versions of the wiki-client older than this don't pick up the factory.json files that need to be added to show their about pages in this menu.

paul90:
While I'm not convinced there are any breaking changes here, or in the associated plug-in changes, I'm incrementing the minor version for these changes. *And will be pulling them all together in an updated wiki version.*

Published as + wiki-client@0.15.0

nrn:
Thanks @paul90! The breaking change isn't a huge deal, but if an older wiki-client pulled in a newer changes or activity plugin, they would show up in the "other" column of the new paragraph factory, and people could accidentally create paragraphs with those types even though they don't really mean anything. Admittedly it should only happen in a kinda weird situation, like a development environment that's only getting updated piecemeal, but it seems like a confusion worth avoiding.

PR #229, opened 2019-04-05
Convert Selected Plugin Pages to references with synopses

nrn:
Paired on this with @wardcunningham

image

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -17,6 +17,8 @@ license = require './license'
 asSlug = require('./page').asSlug
 newPage = require('./page').newPage
 
+wiki.origin.get 'system/factories.json', (error, data) ->
+  window.catalog = data
 
 $ ->
   dialog.emit()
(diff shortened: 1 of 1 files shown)


WardCunningham:
The synopsis of many of these pages refer to the other similar pages. This is leftover from the time when there was one relevant link, Recent Changes, on the default welcome visitors page. I'd like to see the synopsis be more specifically about the page to which it is attached. A good move might be to move these extra links to a second paragraph. Maybe if the hamburger menu becomes the primary navigation mechanism we can forget about cross-linking the various automatic lists of pages.

WardCunningham:
We should have some guidelines for how these pages should be written. We could start by choosing a word to distinguish them. We are referring to them as "Selected Plugin Pages" here and in the user interface. We've also called them "Utility Plugin Pages" somewhere before. Is there a better word?

A utility page should include the following:

- Synopsis that says simply what it does and perhaps what else can be done.
- Related pages can be recommended in a second paragraph, not the synopsis.
- A paragraph of special instructions and consequences, use (x) to remove.
- The generally configured plugin item which should appear distinctive and provide utility.
- A "See About .." link to the plugin's about page, same as cmd/alt-i when editing.

If these seem reasonable, I will prepare pull requests to make the current pages conform.

paul90:
Published as + wiki-client@0.15.1

WardCunningham:
Improved synopsis for Activity, Changes and Recycler now submitted as pull requests.

PR #231, opened 2019-04-07
Set tabindex of pages and focus them when active.

joshuabenuck:
When changing the active page with the left / right arrow keys, the keyboard focus doesn't follow which makes it impossible to scroll the new active page without using the mouse.

This small change enables the page div to be focusable (by setting the tabindex) and then uses this new ability to set the keyboard focus whenever a page is made active.

I chose a tabindex of -1 so pages aren't added to the list of tab selectable elements.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -272,10 +272,11 @@ p.readout {
   background-color: white;
   height: 100%;
   overflow: auto;
-  box-shadow: 2px 1px 4px rgba(0, 0, 0, 0.2); }
+  box-shadow: 2px 1px 4px rgba(0, 0, 0, 0.2);
+  outline: none; }
   .page.active {
     box-shadow: 2px 1px 24px rgba(0, 0, 0, 0.4);
-    z-index: 10; outline: none; }
+    z-index: 10; }
 
 .page.plugin > .paper {
   box-shadow: inset 0px 0px 40px 0px rgba(0,220,0,.5);
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -275,7 +275,7 @@ p.readout {
   box-shadow: 2px 1px 4px rgba(0, 0, 0, 0.2); }
   .page.active {
     box-shadow: 2px 1px 24px rgba(0, 0, 0, 0.4);
-    z-index: 10; }
+    z-index: 10; outline: none; }
 
 .page.plugin > .paper {
   box-shadow: inset 0px 0px 40px 0px rgba(0,220,0,.5);
--- a/lib/active.coffee
+++ b/lib/active.coffee
@@ -33,3 +33,4 @@ active.set = ($page) ->
   $page = $($page)
   $(".active").removeClass("active")
   scrollTo $page.addClass("active")
+  $page.focus()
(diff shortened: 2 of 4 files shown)


WardCunningham:
I will explore this pull request with Joshua, a colleague here at New Relic.

Some info I have found useful:
https://stackoverflow.com/questions/1599660/which-html-elements-can-receive-focus

WardCunningham:
Joshua and I met and experimented with this pull request from localhost. We found a couple of small things that he will investigate:
- Suppress blue focus halo were we already have the subtle gray halo. (css)
- Discover why newly added pages aren't focusable until tab reload. (refresh)

joshuabenuck:
Okay. I have fixed the two problems found while testing with Ward.
1. *Blue halo issue*: Solved it by adding an additional style to remove the outline to the CSS rule we already use to add the drop shadow to active pages. This will restrict the change in styling to just the halo around active pages and will not affect the halo around editors for items within pages. Here's a [useful post](http://usabilitypost.com/2008/10/15/css-tip-remove-mac-osx-glowing-blue-outlin-for-custom-styled-input-fields/) that demonstrates the approach.

   I have verified that we still get a halo when editing items within pages and that the drop shadow is the only halo we get around active pages. I will do one more round of testing on my Windows laptop to double check that things work there as well.

2. *Render issue*: There was an additional way to create pages dynamically which doesn't use the template in static.html. This (createPage) has now been updated to include a tabindex.

   I have verified that the keyboard focus follows the drop shadow in the following situations:
   * tab reload
   * click on an internal link to an existing page
   * click on internal link to a page that doesn't exist
   * click on a previous version of a page
   * dragging a page from another site into the frame
   * clicking on internal links within an external page.

joshuabenuck:
All of the above work on a separate checkout on my Windows laptop. We should be good to go. If there are additional scenarios I should test, please let me know!

paul90:
Thank you Joshua.

I was going to ask _What Blue halo?_, but spotted it when I tried things out with Chrome. The blue halo in Chrome, and other Chromium based browsers, sure is intrusive compared to the more subtle grey dotted outline in Firefox.

I notice that the blue halo is still there on the page while using the mouse to select the wiki page, if only while the mouse button is down. I wonder if we want to replace this with something less intrusive, like the Firefox experience? This could be achieved by adding
``` css
.page:active {
  outline: grey dotted 1px;
}
```
We could also change the halo around the link that has been tabbed to, has focus, to give something less intrusive like the experience in Firefox. Though this might make it difficult to spot for those more use to the blue halo.
``` css
a:focus {
  outline: grey dotted 1px;
}
```
These two css rules could always be combined

joshuabenuck:
The halo issue was less that the blue outline in general was intrusive, but that it was different that what we did before. We already have an outline for pages (the drop shadow) and having another was both redundant and different. I'm inclined to leave the styling for focused links and editors as is for the reason you mentioned, others may be used to it.

The latest commit sets at outline of none for all pages, not just active pages which removes the blue halo on mouse down.

paul90:
Looks good. Merging

paul90:
Published as + wiki-client@0.15.1

PR #232, opened 2019-04-10
Don't propagate editor click events to page.

joshuabenuck:
This fixes an issue caused by PR #231. When you click in an already active editor to reposition the cursor, the click event propagates up to the parent which calls active.set() which causes the active editor to lose the focus when focus is called on the page.

Here are the ways I can see this issue could be fixed. (This PR goes with option 1):
1. Tell the editor to not allow the event to propagate to the parent.
2. In active.set() change the call to only call focus if there is no element with the .textEditing style present in the DOM.
3. In active.set() don't call focus if the to be activated page is the currently active page.

Options 2 and 3 feel like they are trying to undo something that should not have happened. Option 1 is the only one that prevents the problem from happening in the first place.

I've verified that:
   * focus is still following the active pane when using the arrow keys or when clicking with the mouse
   * editors can still be activated by double clicking
   * the cursor can be repositioned in an active editor using the mouse

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -111,6 +111,7 @@ textEditor = ($item, item, option={}) ->
   $textarea = $("<textarea>#{escape original}#{escape option.suffix ? ''}</textarea>")
     .focusout focusoutHandler
     .bind 'keydown', keydownHandler
+    .click (e) -> e.stopPropagation()
   $item.html $textarea
   if option.caret
     setCaretPosition $textarea, option.caret


WardCunningham:
Thanks for following up on this.

It might be worth looking at these files to see what they are doing with click. Github will find the file but only shows you one of the occurences of the search term so you have to go look.

https://github.com/fedwiki/wiki-client/search?q=click&unscoped_q=click

WardCunningham:
Incidentally, there are several overviews of the codebase which might help you find your way around.

http://stack.fed.wiki -- highest level
http://dev.asia.wiki.org -- code walkthroughs

joshuabenuck:
Thanks! While trying to figure out a way to fix this I logged a stack trace every time active.set was called to double check that I hadn't just missed a call path. I'll spend some time looking through the click handlers to see if any could be adversely impacted by my proposed change.

Thanks for the links to the overviews! I was not aware of them.

joshuabenuck:
Okay. I've completed looking through references to click and double click in the wiki-client codebase. It did not turn up anything that I feel will be adversely impacted by my proposed change.

I did find one new alternative way to implement my fix.

There is a click handler for elements with the page class.
```coffeescript
    $('.main').delegate '.page', 'click', (e) ->
       active.set this unless $(e.target).is("a")
```
Since we are already employing a guard clause, we could just add another:
```coffeescript
    $('.main').delegate '.page', 'click', (e) ->
       active.set this unless $(e.target).is("a") or $(".textEditing").length != 0
```
My preference still is to prevent the event from propagating rather than to add an additional guard clause in the click handler.

As an aside: If I knew why the current guard clause was in place, I'd try stopping the propagation on link clicks instead. I tried commenting out the unless clause, but wasn't able to get anything to fail. The guard clause has been there since the code was ported from SFW and I wasn't able to find the original source file in SFW to dig further.

My recommendation is to move forward with the fix I've currently proposed. Please let me know if there is any additional research or testing I need to do.

Research notes:
   * The image plugin is another place where we use stopPropagation in the codebase. It is not immediately apparent to me what behavior we are preventing, but it does show precedence in the codebase.
   * The factory plugin calls active.set for its hosting page, but does so before creating its editor so it doesn't cause any problems with the editor focus.

paul90:
> The guard clause has been there since the code was ported from SFW and I wasn't able to find the original source file in SFW to dig further.

It gets tricky, you need to find the commit just before the client was removed from there to be replaced with adding it using NPM. 

It was originally added to `client.coffee` back in October 2011, in [commit 3881423](https://github.com/WardCunningham/Smallest-Federated-Wiki/blame/3881423169aa1b4dbbfc559a12204d80701717f2/client/client.coffee#L362-L363)

WardCunningham:
The original source was all in what we now call a "mono repo" which is trendy again.

https://github.com/WardCunningham/Smallest-Federated-Wik
It's ReadMe warns:i
```
This repository exists as both a historical document and 
a community of interested parties. This is not where you
want to find the current source for Federated Wiki.
```

Here is some video history, thanks to Paul.
http://video.fed.wiki.org/sfw-history-animated.html

paul90:
Published as + wiki-client@0.15.2

PR #234, opened 2019-04-14
Prevent flicker and loss of focus from wiki page elements during lineup scrolling

### LINKED ISSUE #233: Focus and flicker issue caused by recent PRs
I have found two additional issues caused by my recent changes (#231 and #232).
1. It is impossible to get the Federation Search to retain focus. This has the same root cause as #232 and points to a flaw in how that was originally fixed.
2. There's a noticeable flicker when scrolling back through to previous pages. I suspect the call to focus in #231 is what is causing this, but haven't dug deep enough to prove it or think of a fix.

You can see both [here](https://server.wiki.randombits.xyz/view/welcome-visitors/view/federation-search/view/dns/view/hosting-options). Try clicking in the Federation Search box or using the arrow keys to scroll back to the page hidden on the left. The latter is easiest to see if you hit left and right a few times. Neither issue is present if I rollback my changes.

What is the best way for us to proceed here? I'm inclined to roll back the changes until all of these issues are fully understood, fully fixed, and the fixes have had some time to bake, but I'm also willing to continue to push forward from where we are now.

paul90:
Fixes #233 

Changes to how we set page focus, so that it is only done after the page scrolling animation is complete (to prevent flicker), and if focus is not already set within the current wiki page.

### DIFF
--- a/lib/active.coffee
+++ b/lib/active.coffee
@@ -32,7 +32,9 @@ scrollTo = ($page) ->
   # scroll to target and set focus once animation is complete
   active.scrollContainer.animate({
     scrollLeft: scrollTarget
-    }, () -> $page.focus())
+    }, () ->
+      # only set focus if focus is not already within the page to get focus
+      $page.focus() unless $.contains $page[0], $( document.activeElement ))
 
 
 active.set = ($page) ->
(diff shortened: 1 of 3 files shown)


WardCunningham:
I'm sorry. I missed your request for review. I just now read the generated email. I'll look closely at these changes before the day is out.

paul90:
With the change in d2b2efe we don't need to modify plugins with input fields, so closing the related pr in html and search plugin.

paul90:
Published as + wiki-client@0.15.4

PR #235, opened 2019-04-28
Drag to reorder pages

joshuabenuck:
This PR enables pages to be reorder by dragging them. The URL is updated to the new lineup once a page has been dropped.

A couple of implementation notes:
* I was unable to find a way to prevent the user from dragging infinitely to the right. After the first checkin, it was possible to be stranded on the far right without any way to scroll back to content. The second checkin fixes that by scrolling the dropped page back into view.
* A new parameter was added to active.set to allow the active page to be set without immediately scrolling to it. This prevents some odd scrolling back and forth scrolling behaviors in longer lineups where pages are partially obscured.

A desired future addition is to allow pages to be removed by swiping up or down. That may change some of the implementation details here, but I feel the current changes provide value as-is.

### DIFF
--- a/lib/lineup.coffee
+++ b/lib/lineup.coffee
@@ -18,6 +18,12 @@ addPage = (pageObject) ->
   keyByIndex.push key
   return key
 
+changePageIndex = (key, newIndex) ->
+  oldIndex = keyByIndex.indexOf key
+  page = keyByIndex[oldIndex]
+  keyByIndex.splice(oldIndex, 1)
+  keyByIndex.splice(newIndex, 0, page)
+
 removeKey = (key) ->
   return null unless key in keyByIndex
   keyByIndex = keyByIndex.filter (each) -> key != each
@@ -86,4 +92,4 @@ crumbs = (key, location) ->
   result
 
 
-module.exports = {addPage, removeKey, removeAllAfterKey, atKey, titleAtKey, bestTitle, debugKeys, debugReset, crumbs, debugSelfCheck}
+module.exports = {addPage, changePageIndex, removeKey, removeAllAfterKey, atKey, titleAtKey, bestTitle, debugKeys, debugReset, crumbs, debugSelfCheck}
--- a/lib/lineup.coffee
+++ b/lib/lineup.coffee
@@ -20,9 +20,8 @@ addPage = (pageObject) ->
 
 changePageIndex = (key, newIndex) ->
   oldIndex = keyByIndex.indexOf key
-  page = keyByIndex[oldIndex]
   keyByIndex.splice(oldIndex, 1)
-  keyByIndex.splice(newIndex, 0, page)
+  keyByIndex.splice(newIndex, 0, key)
 
 removeKey = (key) ->
   return null unless key in keyByIndex
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -102,9 +102,6 @@ header {
 .header > h1 {
   cursor: grab; }
 
-.ui-sortable-helper .header > h1 {
-  cursor: grabbing; }
-
 footer {
   bottom: 0; }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -102,6 +102,9 @@ header {
 .header > h1 {
   cursor: grab; }
 
+.ui-sortable-helper .header > h1 {
+  cursor: grabbing; }
+
 footer {
   bottom: 0; }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -313,6 +313,19 @@ p.readout {
   top: 0px;
 }
 
+div.page-draggable.page {
+  box-shadow: 2px 1px 24px rgba(255, 250, 153, 0.4);
+  transition-property: box-shadow;
+  transition-duration: 300ms;
+}
+
+/* make more specific than ghost style. */
+div.pending-remove.page {
+  opacity: 0.2;
+  transition-property: opacity;
+  transition-duration: 300ms;
+}
+
 .twins span {
   z-index: 2;
   position: relative;
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -99,7 +99,7 @@ img.remote,
 header {
   top: 0; }
 
-.header span {
+.page-handle {
   cursor: grab; }
 
 footer {
@@ -116,14 +116,6 @@ footer {
 .footer:hover {
   opacity: 1;
 }
-
-.header > h1 {
-  display: table; }
-
-.header > h1 > span {
-  width: 100%;
-  display: table-cell;
-  padding-left: 5px; }
 
 .story {
   padding-bottom: 5px; }
@@ -309,6 +301,26 @@ p.readout {
   /*width: 430px;*/
   padding: 30px;
   min-height: 100%;
+  z-index: 0;
+  position: relative;
+}
+
+.page-handle {
+  z-index: 1;
+  width: 100%;
+  position: absolute;
+  left: 0px;
+  top: 0px;
+}
+
+.twins span {
+  z-index: 2;
+  position: relative;
+}
+
+.header span {
+  z-index: 2;
+  position: relative;
 }
 
 .factory,
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -117,6 +117,14 @@ footer {
   opacity: 1;
 }
 
+.header > h1 {
+  display: table; }
+
+.header > h1 > span {
+  width: 100%;
+  display: table-cell;
+  padding-left: 5px; }
+
 .story {
   padding-bottom: 5px; }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -99,6 +99,9 @@ img.remote,
 header {
   top: 0; }
 
+.header > h1 {
+  cursor: grab; }
+
 footer {
   bottom: 0; }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -99,7 +99,7 @@ img.remote,
 header {
   top: 0; }
 
-.header > h1 {
+.header span {
   cursor: grab; }
 
 footer {
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -301,6 +301,9 @@ p.readout {
   /*width: 430px;*/
   padding: 30px;
   min-height: 100%;
+}
+
+.handle-parent {
   z-index: 0;
   position: relative;
 }
--- a/lib/active.coffee
+++ b/lib/active.coffee
@@ -37,7 +37,8 @@ scrollTo = ($page) ->
       $page.focus() unless $.contains $page[0], document.activeElement )
 
 
-active.set = ($page) ->
+active.set = ($page, noScroll) ->
   $page = $($page)
   $(".active").removeClass("active")
-  scrollTo $page.addClass("active")
+  $page.addClass("active")
+  scrollTo $page unless noScroll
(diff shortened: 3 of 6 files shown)


joshuabenuck:
Enabling drag prevents you from being able to select text. One of my use patterns is to select text off of a page when pasting wiki content into external programs (this requires you to be out of edit mode). This breaks that workflow.

How should this be fixed? Only allow page drag when a key is held down (or vice versa)? Tie it to edit mode (which might cause confusion since page drag is a reordering operation and not an edit operation)? Restrict the draggable element to be the page header?

I'm inclined to make it key based and change the mouse cursor to indicate that you are in page reordering mode.

Thoughts?

joshuabenuck:
Don't merge yet. While the cursor still changes to the pointer when hovering over the page header flag, you are no longer able to drag the flag.

joshuabenuck:
Here's where things stand now:
* Both text selection and flag dragging work as they did before.
* Dragging a page can only be done by dragging the page header text. A span was added to the page header to give us something to bind to that is separate from the header. This allows both flag dragging and page dragging to work.
* The user is informed of the ability to drag a page by the mouse cursor changing to the _grab_ cursor when hovering over the page title text. It switches to a _grabbing_ cursor when the user initiates a drag and remains the _grabbing_ cursor until they let go whether or not the mouse cursor remains over the page title text.
* The URL is updated once the user drops the page. This uses the existing _state.setUrl_ call which means that the back button will undo the reorder operation.

I wasn't able to come up with a CSS selector to pass to sortable that included the text, but excluded the flag link and its image. This is why the new span was added.

I do not know the codebase well enough to say for certain that adding the span around the header's h1 text content breaks anything.

**Important note:** The cursor will not change to the _grabbing_ cursor with Chrome dev tools open! I lost a lot of time chasing my tail because of this.

You can try out the changes at http://fed.wiki.randombits.xyz/

joshuabenuck:
The latest commit removes the axis movement restriction in order to allow pages to be dragged up off of the page. If dropped there, the page is removed from the lineup. The changes are on randombits if you want to try them out.

A few points of note:
* The item that occupies the space in the lineup of the removed page is the page that receives focus after the removal unless the removed page was the last in the lineup.
* Users are prevented from removing the last item in the lineup. (As an aside: I feel we should also prevent hitting the browser back button from removing the last item in a lineup. Today, the last item can be removed by hitting back).
* Dragging down, to the left, or to the right will not allow a page to be removed.
* Removal respects the browser history as it follows the same code path as reordering so you can use the back button to restore the page you removed.

Give it a try and let me know what you think.

WardCunningham:
Many improvements. I was going to suggest the top 1/4 inch of the page as the drag handle but the Title seems to work well.

joshuabenuck:
The latest commit expands the header span to fill the available space. This makes it easier to reorder or remove pages that have short titles. Previously, you had to grab the text directly. Randombits has been updated with the changes.

WardCunningham:
In the original design we adopted the then popular architecture of displaying content retrieved from the server and then returning deltas back to the server in response to visual changes.
```
  page json => .page div
  jquery event => server action => .journal update
```
Within a few years common practice included the parallel maintenance of a client-side model. We've partially adopted this approach which was intended to be the true state of the client. Relevant to this work would be the model known as the lineup. [source](https://github.com/fedwiki/wiki-client/blob/master/lib/lineup.coffee)
```
  page json => .page div with key
            => lineup model with keyed and ordered list of pages
```
The keys in the dom connect .page divs there with the pages in the lineup model. These can get out of sync at which point operations might or might not work correctly depending on whether they trust the dom or the model.

We report alignment to the javascript console whenever operations change either order.

image
(This from Firefox. Chrome no longer shows the keys themselves when logged.)

This is a concern that will require considerable attention if it is ever to be bug free. A good next step would be to study how various event handlers revise both and pay particular attention to possible existing flaws lurking in this logic. The reordering handlers will of course need to duplicate this logic also. A quick look this morning shows they get quicky out of sync.

Here I reordered one page and then click a link on the new rightmost page to add another.

image

Some complications that needed to be addressed include making a distinction between a page and a div representing that page. The server has only one page but the client can have multiple copies in the lineup and they need not all be the same. A common workflow has one fork an earlier version of a page and then drag content from the more complete but now client-side only version to the now active page.

Some benefits we sought from introducing the lineup involved considering it as an abstract representation of a computation that could be further manipulated, evaluated and transformed. Much of this remains an aspiration.

WardCunningham:
I particularly like the way unwanted pages can be removed from the lineup by dragging them up and away. There were a couple of occasions where I removed a page I intended only to reorder. I wonder if I could be given advance warning of the distinction while the drag is in progress?

Perhaps a page could become transparent when a drag-release would make it vanish. Ghost pages are already a bit transparent so the desired effect would need to be more dramatic. Could they be made barely visible with the transition animated over a few hundred milliseconds?

joshuabenuck:
The visual changes you are recommending would definitely help. I was also thinking of requiring the header text be dragged off before triggering a remove (or least more of the whitespace above the header). Right now a remove is triggered if any part of the page is off screen when the drop happens.

I'll get those implemented first and will then dig into the model problems that have been introduced.

joshuabenuck:
The title now has to be dragged out of the window in order to trigger the remove action which requires a lot more intention than it did before. The page is also faded out to 20% opacity to communicate to the user what will happen on drop. The changes are live on fed.wiki.randombits.xyz.

joshuabenuck:
The lineup and the dom should now stay in sync. I had to add an extra method to lineup.

I extracted out the console logs from state.show into their own method so they can be called on sortstop to ease troubleshooting.

The implementation of state.show brings up an interesting consideration. It assumes that between the old state and the new state that all commonalities are at the beginning. URL hacking already violates these assumptions, but that was the exception. With these changes, it will become far more common.

Consequently, browser history navigations between states where there have been any non-linear lineup changes will result in more pages being reloaded than strictly necessary. When something, for example, has been added to the beginning of the lineup, the entire lineup will be reloaded.

It may be time to revisit this implementation if we care about minimizing page reloads. It's purely a performance / aesthetics concern for static pages. Pages with plugins are another story, one which I haven't dug into yet.

The changes are deployed on randombits for those that want to try them out.

As an aside: a refactored state.show that is smarter and also doesn't force focus the last item in the lineup might allow us to call it instead of making the surgical updates in the sortstop event handler.

joshuabenuck:
With the ability to reorder the page lineup and remove pages, I find myself wanting to be able to open pages next to the focused page and not just have them open at the end of the lineup.

We already use shift click to open multiple pages. What if we use Ctrl/Cmd click to open a page immediately to the right of the focused page?

WardCunningham:
When the browser emits [popstate](https://github.com/fedwiki/wiki-client/blob/6c3959d892aae8a2b669932e3b7a77de8feae233/lib/legacy.coffee#L45) we should first try to discover what ui interaction might have been reversed and apply those operations with the already instantiated lineup. For example, cases where we have the name number of each site-slug then we should look for the one operation that explains this. If there are more than one possibility, we could choose the one that spans the shortest distances or are most visible in the current horizontal scroll.

Similarly when only one site-slug changes and it changes by + or - one, then we still have a lot to go on. The guesswork here is because the browser state is incomplete compared to what we have stored in the lineup model with all those unique keys. During the actual move operation we should know exactly what happened and not need any heuristic.

For a highly ambiguous case try singing a few more verses of the bottles of beer song [here](http://start.fed.wiki/view/seventh-floor) and then try a few reorder operations. If you have 97, 96, 95, 94 bottles of beer and you reorder one page would you expect to see 97, 95, 96, 94? You could argue yes by the "persistence of view" argument. But we've discussed the ideal of the "consistent interpretation of configuration" argument too and it says the answer must be 97, 96, 95, 94.

lib/lineup.coffee
```diff
@@ -18,6 +18,12 @@ addPage = (pageObject) ->
   keyByIndex.push key
   return key
 
+changePageIndex = (key, newIndex) ->
+  oldIndex = keyByIndex.indexOf key
+  page = keyByIndex[oldIndex]
+  keyByIndex.splice(oldIndex, 1)
+  keyByIndex.splice(newIndex, 0, page)
```
nrn: This method looks right, but you don't need to get `page` out of the keyByIndex, since it will always be the same as `key` that is passed into this function.

nrn:
Playing around with the example it's really nice to use! The one weirdness I noticed is that if you try and remove the last open page, it doesn't remove it (which makes sense to me), but it does stay mostly transparent.

joshuabenuck:
Good catch! Since the last page is not removable, the reduction in opacity shouldn't even be applied in the first place.

lib/lineup.coffee
joshuabenuck: Thanks for calling this out. The change has been made.

joshuabenuck:
The last page in a lineup will no longer be marked for removal which resolve's the issue Nick noticed. Changes are live on randombits.

nrn:
Looks great!

joshuabenuck:
The page grab handle is now the area above the story with the exception of the twins and header. There's now an added visual cue of the page fading by 20% when it can be grabbed. This also allows one to see behind the page being grabbed during reordering which may help you choose where to drop it.

Changes are live on randombits. Please give them a try and let me know what you think. Is the larger grab area better? The fading too much, not enough, not needed? Is it the wrong type of visual cue?

These changes open up the possibility for using click or double click as a way to rename the page. The previous implementation would have made the UX of such a rename operation awkward.

joshuabenuck:
This breaks the opacity on ghost pages. I'm going to switch to applying and removing a class instead of using an inline style which should fix that issue. I may also try using a drop shadow instead of changing opacity.

joshuabenuck:
Using a drop shadow works better with ghost pages. Switching to CSS3 transitions instead of jquery animations fixed the problem with ghost pages losing their translucency when moving off the page and back on. Feedback welcome. Changes are live on randombits.

WardCunningham:
I'm thinking we're ready to merge and release this code. Do you agree, @joshuabenuck?

joshuabenuck:
@WardCunningham - I agree (assuming you've had a chance to try out the change to the page drag handle / the page highlight that shows up when a page is draggable). I've been surprised at how often I go to grab the page header, but I suspect that is because I was used to the original implementation.

joshuabenuck:
If you drag an item from a page on the left to a page on the right, the page on the left starts to scroll. This doesn't happen on master. Please hold off on merging until I figure out why. This issue significantly impacts the usability of dragging between pages.

joshuabenuck:
OK. Things are looking stable. Feel free to merge and release once you are ready.

paul90:
@joshuabenuck - Just spotted a problem...

With Firefox - If while dragging you release the mouse button in a location that does not result in the page order changing - the page retains the coloured shadow.

paul90:
A solution is to remove the `page-draggable` class when on `sortstop`.

But, I am tempted to simply remove the change in shadow colour - as even removing at the end of the sort does not seem to behave consistently either across browser to an repeated use.

paul90:
In 185c175b0f24fae2ea8914ff3223e1a865aba62d I've removed `page-draggable`.

joshuabenuck:
Good catch. Looks like I need to increase the number of browsers I use in testing. I've installed FireFox now and can reproduce what you were seeing. I think removing page-draggable was the right thing to do. I've been considering doing a translucent overlay for page-draggable. I think it better communicates the action the user can take. Either way, that's for another PR. Thanks for getting this merged!

paul90:

PR #236, opened 2019-05-01
Rework copy and move item user experience.

joshuabenuck:
This PR changes the mouse cursor when reordering items within a page or moving / copying between pages based. They also give the user some visual feedback when performing a copy.

These changes likely need some burn in time before they can be committed. I'm starting the PR now to get feedback and gives others the opportunity to try them out. I don't have them hosted anywhere at the moment. I may need to spend some time tweaking my hosting setup so I can host different wikis running different versions of the codebase. I'm limited to one-ish right now.

Changes:
* Use the copy cursor when shift is held down.
* When the copy cursor is shown, show the original item on the source page.
* Use the move cursor by default.
* Do not show the copy cursor when on source page.
* Tweak mouse offset on dragged items.

This does not add a "copy item" entry into the journal. I'm not even sure we have that concept. If needed, I can investigate adding that.

I'm also tempted to rework the code a bit more as the sortable API already has callbacks for some of the states we are manually computing. It may make the code a bit easier to follow.

I have not tested the move or copy semantics related to ghost pages that I see in the code. I may need someone to do this or walk me through the intent behind that code path.

There are probably other things I'm missing.

### DIFF
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -40,6 +40,8 @@ aliasItem = ($page, $item, oldItem) ->
     item.alias ||= item.id
     item.id = random.itemId()
     $item.attr 'data-id', item.id
+    $item.data('id', item.id)
+    $item.data('item').id = item.id
   else if item.alias?
     unless pageObject.getItem(item.alias)?
       item.id = item.alias
@@ -49,28 +51,36 @@ aliasItem = ($page, $item, oldItem) ->
 
 handleDragging = (evt, ui) ->
   $item = ui.item
+  $item.removeClass('copy-only copy-capable')
 
   item = getItem($item)
   $thisPage = $(this).parents('.page:first')
   $sourcePage = $item.data('pageElement')
+  sourceIsGhost = $sourcePage.hasClass('ghost')
   sourceSite = $sourcePage.data('site')
 
   $destinationPage = $item.parents('.page:first')
+  destinationIsGhost = $destinationPage.hasClass('ghost')
   equals = (a, b) -> a and b and a.get(0) == b.get(0)
 
   moveWithinPage = not $sourcePage or equals($sourcePage, $destinationPage)
   moveFromPage = not moveWithinPage and equals($thisPage, $sourcePage)
   moveToPage = not moveWithinPage and equals($thisPage, $destinationPage)
 
-  # If making a copy, update the temp clone so it becomes a true copy.
-  if ui.item.hasClass('copy')
-    $('.shadow-copy').removeClass('shadow-copy')
-      .data(ui.item.data()).attr({'data-id': ui.item.attr('data-id')})
+  if destinationIsGhost
+    $(evt.target).sortable('cancel')
+    $('.shadow-copy').remove()
+    return
 
   if moveFromPage
-    if $sourcePage.hasClass('ghost') or
-      $sourcePage.attr('id') == $destinationPage.attr('id') or
-        evt.shiftKey
+    # If making a copy, update the temp clone so it becomes a true copy.
+    if ui.item.hasClass('copy')
+      $('.shadow-copy').removeClass('shadow-copy')
+        .data(ui.item.data()).attr({'data-id': ui.item.attr('data-id')})
+      return
+
+    if sourceIsGhost or
+      $sourcePage.attr('id') == $destinationPage.attr('id')
           # stem the damage, better ideas here:
           # http://stackoverflow.com/questions/3916089/jquery-ui-sortables-connect-lists-copy-items
           return
@@ -91,7 +101,13 @@ handleDragging = (evt, ui) ->
   pageHandler.put $thisPage, action
 
 changeMouseCursor = (e, ui) ->
-  if e.shiftKey and ui.item.hasClass('copy-capable')
+  destinationIsGhost = ui.placeholder.parents('.page:first').hasClass('ghost')
+  if destinationIsGhost
+    $('body').css('cursor', 'no-drop')
+    $('.shadow-copy').hide()
+    ui.item.removeClass('copy')
+  else if ui.item.hasClass('copy-only') or \
+    (e.shiftKey and ui.item.hasClass('copy-capable'))
     $('body').css('cursor', 'copy')
     $('.shadow-copy').show()
     ui.item.addClass('copy')
@@ -122,7 +138,10 @@ initDragging = ($page) ->
           zIndex: ''
         ).removeAttr('data-id')
     .on 'sortover', (e, ui) ->
-      if ui.sender and ui.sender[0] != e.target
+      sourceIsGhost = ui.item.data('pageElement').hasClass('ghost')
+      if sourceIsGhost
+        ui.item.addClass('copy-only')
+      else if ui.sender and ui.sender[0] != e.target
         ui.item.addClass('copy-capable')
       else
         ui.item.removeClass('copy-capable')
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -122,7 +122,7 @@ initDragging = ($page) ->
           zIndex: ''
         ).removeAttr('data-id')
     .on 'sortover', (e, ui) ->
-      if ui.sender
+      if ui.sender and ui.sender[0] != e.target
         ui.item.addClass('copy-capable')
       else
         ui.item.removeClass('copy-capable')
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -62,6 +62,11 @@ handleDragging = (evt, ui) ->
   moveFromPage = not moveWithinPage and equals($thisPage, $sourcePage)
   moveToPage = not moveWithinPage and equals($thisPage, $destinationPage)
 
+  # If making a copy, update the temp clone so it becomes a true copy.
+  if ui.item.hasClass('copy')
+    $('.shadow-copy').removeClass('shadow-copy')
+      .data(ui.item.data()).attr({'data-id': ui.item.attr('data-id')})
+
   if moveFromPage
     if $sourcePage.hasClass('ghost') or
       $sourcePage.attr('id') == $destinationPage.attr('id') or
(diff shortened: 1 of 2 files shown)


WardCunningham:
We haven't distinguished between a copy or a move. Both are type 'add' as is the product of the [+] button which creates and adds a new item to the page. This is consistent with a very local perspective of the journal which must make sense in whatever context it might appear.

joshuabenuck:
Thanks for the clarification. I went ahead and put up a wiki running this PR so it is more easily tested [dev.fed.wiki.randombits.xyz](http://dev.fed.wiki.randombits.xyz). (Note: This is a separate instance from the one running the reordering changes.)

paul90:
I think the only issue I'm seeing is that when you shift+drag an item off the page if you then move back onto the page the item was on visually it looks as if you are still making a copy. When dropped back onto the original page you don't get a copy.

@WardCunningham A deeper question, I guess, is should shift+drag within a page behave as now and just move the item or should it make a copy of the item.

joshuabenuck:
Interesting find. It looks like the copy-capable class is never being removed once it is added. This means that you can get the same behavior you noticed just by dragging an item over a different page and then back to the original.

joshuabenuck:
The latest change fixes the mouse cursor issue when hovering back over the original page, but I'm not seeing the copy persist after refresh. I need to dig into this a bit more.

joshuabenuck:
Quick update on this. I must have somehow corrupted the state of my local wiki changes. Drag and drop behavior got really weird on me until I delete all local changes after which the results of many of my previous changes were suddenly there. More testing is needed to see if I can reproduce this.

WardCunningham:
@paul90, from the user perspective drag should move unless copy is called for by circumstance. For example, drag from a ghost page is always a copy. Shift-drag should always be a copy.

I think this is pretty much how it works now except the sensing for shift might occur once when it would be better if it were continuous throughout the drag.

There are some visual queues that we don't have and are possibly not available from jQuery UI. Particularly, when the drag is a copy it should not remove the original from its visual location. If alternating between move and copy in the midst of a drag then we could wish for the original item to appear and disappear as the shift is pressed and released.

No action should take place until the drag completes.

We have had situations where a copy or even a deep copy was required to avoid some javascript circularity. As long as the correct actions are created with the correct ids then all is good.

A move between pages will generate two actions, an add and a delete. I vaguely remember that this order of operations was important in some degenerate case. Best not to change it.

joshuabenuck:
@WardCunningham This PR will cause the original item to appear or disappear based on whether shift is held down (even mid drag), but it doesn't currently allow creating a copy within the same page. You have to drag it to another page in order for copy to be an allowed option. Is that something I should change?

Dragging from a ghost page currently looks like it is a move. I need to do more work to fix that workflow.

joshuabenuck:
Drag and drop within ghost pages as well as across duplicate copies of a single page are now prevented. The mouse cursor is updated based on the current operation. You may have to move the mouse a bit after holding down or releasing shift to get the cursor to change. Note that on some OS's the mouse cursor doesn't properly update if the browser dev tools are open.

After walking through 1f52d5a with Ward, I went back through and cleaned up / simplified several things in the implementation.

Test cases I've been using:
* Move within a single page.
* Attempt to copy within a single page (cursor won't change).
* Move across two different pages.
* Copy across two different pages.
* Attempt to move or copy within a ghost page (opened from the journal).
* Open two copies of a single page. Attempt to move and copy between them. Both should be prevented.
* Open an old copy of a page. Copy an item from the previous version that exists in the new version. Move the item after copying. Refresh the page. The moved item should be in the new location and hovering over the last journal entry while holding shift should highlight that moved item. (There was a bug where the new location would be lost on refresh and the journal entry would reference the other copy of the item on the page).

These changes are live on fed.wiki.randombits.xyz for anyone that wants to try them out.

WardCunningham:
@joshuabenuck has found his way into dusty corners of our codebase and completed improvements that have stumped us all these years. I appreciate how he is willing to match the style of the code he modifies and where there are more than one style present, choose the best to mimic.

joshuabenuck:
FYI: The drag and drop implementation doesn't properly update the page model to reflect the changes made by dragging items over to the page. Upon further investigation, I see that the page model is currently only structured to support initial loading and iterating over the items within the page. It doesn't have the APIs needed to incrementally update the model of the items on the page.

At the moment, the only concrete problem this causes is that aliasItem will not randomize item ids for any item that is copied over since the last time the page was refreshed (or it will randomize the ids for items that were previously on the page, but had been temporarily moved to a different page).

Test case: Copy an item over from a different page (the item copied must not already exist on the page). Then copy the item again. Notice that the data-id attribute for both is the same. Now refresh the page and copy the same item over again. Notice that the data-id for the new copy is randomized.

joshuabenuck:
@WardCunningham Thanks for the comments!

I've started an experiment which wraps all of the state and callbacks used by the item drag operations into a class. It is a departure from the original style, but the added abstraction allows us to more clearly communicate what is happening and reduces the number of times the same or similar variables have to be declared in the callbacks.

This also reduces the complexity of the logic in the drop callback (called stop in the current prototype). Previously, this method was called twice in order to handle updating the source page and the destination page. Now, it is called once which allows the logic to be simplified. The use of a 'copy' class was also eliminated as that state can be stored within the object. Though, these latter two improvements could be integrated into the original style implementation without too much effort.

I was testing these changes when I discovered the problem with the page model.

Here's a gist with the relevant bits. It still needs some cleanup, but is largely complete. https://gist.github.com/joshuabenuck/6dd9277df071cf9edfe0b74f7c06db27

Over time, I'd like to see the code move towards web components which would allow domain operations to live on the elements themselves. A key benefit of that approach is more abstraction / encapsulation of implementation details. Introducing a class here is a small step in that direction.

Anyway, let me know if you feel the approach in the gist is acceptable. If so, I'll clean it up and get it pushed onto this PR.

WardCunningham:
We haven't used coffeescipt class abstraction. Coffeescript offered some class abstraction before ES6 but my committee friends discouraged me from using it. Now coffeescript supports ES6 but in which versions? And do we use those versions?

I'd be more interested in pushing on correctness in the models and mechanisms we do have. Seems like you have identified work to do there. If it is currently not more broken than ever, might still be good to merge this commit before starting another complex project?

joshuabenuck:
Sounds good. I'll abandon the class based approach.

I just discovered that moving an item back to its original location records an unnecessary move in the journal. That is the only problem I know of introduced by these changes. I'll fix this issue and do some more testing. I feel this is close to being ready to merge into master.

joshuabenuck:
handleDrop is now called on sortstop which means it is only called once. I've updated the internals to account for this and I feel it simplifies the logic.

This checkin also removes the need for the 'copy' class.

I still need to track down the source of the unnecessary move operation being added to the journal.

joshuabenuck:
So I see what is going on, but could use some guidance on how to best proceed.

The original implementation didn't mess with the DOM to show / hide the original item during a copy and it also listened to the update event instead of the stop event.

Both of these are leading to the addition of the unnecessary move entry in the journal.

The addition of our shadow-copy element seems to make jQuery think the DOM position has changed when it really hasn't which leads to the update callback being called when it wasn't before. The use of stop instead of update is problematic because even if we didn't have the extra element in the DOM, the stop callback always gets called.

I don't see a way to convince [jQuery UI sortable](https://github.com/jquery/jquery-ui/blob/master/ui/widgets/sortable.js#L1495) to ignore our shadow-copy element. Short of swapping to a different drag a drop implementation (which isn't something I'm inclined to do right now as it doesn't provide much value over what we already have), we could fix this by making our page / item model smart enough to ignore move operations that don't change the order. This feels in line with the enhancements that need to be made to resolve the duplicate data-id issue discovered earlier.

In summary, this PR does make things worse than they are today, but in a fairly subtle way. My recommendation is to live with this flaw and fix it in another PR by making our client side smart enough to ignore no-op changes. One thing I could add to this PR (and have been meaning to do before now) is to introduce a delay on drag start so accidental drags are less common.

I feel the improvements in the user experience and clarity / simplicity of the code changes outweigh the temporary annoyance of unneeded move operations being recorded in the journal if an item is accidentally moved.

Thoughts?

Note that a rebase and force push has been done so the files diff view of the PR now only shows the changes related to this PR. The latest changes are also live on fed.wiki.randombits.xyz.

joshuabenuck:
The delay option has been [deprecated](https://jqueryui.com/upgrade-guide/1.12/#deprecated-distance-and-delay-options). The reasoning behind the deprecation is that interactions should be immediate and that a proper fix is to add drag handles. I plan on sticking with a delay for now. Drag handles have the potential to clutter the UI and / or make eventual touch support more difficult.

The changes are up on randombits. I'm going to turn my attention to the client side item model API. I'll submit those changes as a separate PR.

joshuabenuck:
Move actions are now only recorded if the previous order and current order are different. This is accomplished by storing the order in the story at the beginning of the sort operation and then comparing it with the current order before deciding whether to call pagehandler.put.

When we have a page item model that is kept up to date with edit operations, we can stop storing and removing the original 'order' data in the story element.

The latest change is up on http://fed.wiki.randombits.xyz. I am ready for this to be merged. Give it a try and let me know if you find any issues.

lib/refresh.coffee
```diff
@@ -40,58 +40,105 @@ aliasItem = ($page, $item, oldItem) ->
...
     placeholder: 'item-placeholder'
     forcePlaceholderSize: true
+    delay: 150
   $story = $page.find('.story')
-  $story.sortable(options).on('sortupdate', handleDragging)
+  $story.sortable(options)
+    .on 'sortstart', (e, ui) ->
+      $item = ui.item
+      $story = $item.parents '.story:first'
+      $story.data 'order', getStoryItemOrder($story)
```
nrn: Wold it work to make this a local variable above line 121 instead of storing this order on the element? It would then be passed into handleDrop (which seems like it would be better to call in the same 'sortstop' call as everything else instead of having two).

lib/refresh.coffee
```diff
@@ -40,58 +40,105 @@ aliasItem = ($page, $item, oldItem) ->
...
+      # Create a copy that we control since sortable removes theirs too early.
```
nrn: This adds come complexity it would be cool to avoid if we can. What do we need after the sortable copy is removed?

lib/refresh.coffee
joshuabenuck: The second sortstop is needed in order to avoid more complex logic in handleDrop. The shadow-copy needs to be removed in all cases other than a copy. To do that is a single method would require us to add the same line before every return in handleDrop. Having a second sortstop handler is simpler / results in fewer lines of code.

I'll try putting the original story order in a local variable. If that works, we could safely avoid doing anything on sortstop. I put the cleanup here in order to try to avoid the chance of something else referencing it after the drop. Using a closure would avoid that concern.

lib/refresh.coffee
joshuabenuck: Once we have an always up to date model for items in the page, we won't need to manually save off the order of items on the page here. We would instead get the order from the page model during the drop operation and compare it to the one we are considering.

I'm working on that next, but wanted to get this PR wrapped up to reduce its scope. This is a short term fix to prevent unnecessary moves from being constantly recorded in the journal whenever an item is picked up, but its final location doesn't change.

lib/refresh.coffee
joshuabenuck: Commit 10048c4 switches to using a closure to pass the original order to handleDrop.

nrn: Looks good.

I agree you shouldn't try to overload handleDrop like that, I meant that you could move the `handleDrop(e, ui, originalOrder)` call down to the top of the next `sortstop` handler. It wouldn't change how anything works, but it makes it more explicit that the handleDrop call needs to happen just before the shadow is removed, and makes it less likely for someone to mess it up when working on it later than having it in a separate handler.

lib/refresh.coffee
nrn: Sorry, I think I may still be a little confused here. I meant to ask why we need to do our own `.clone()` below instead of using the sortable element that we were before?

lib/refresh.coffee
joshuabenuck: Ah. That makes sense. 8e5e38e

lib/refresh.coffee
joshuabenuck: We are still using the original sortable element. That's what gets dragged around. What we are doing here is creating a copy of the dragged element to be left behind to represent the original element. This starts out hidden and is shown when we switch from doing a 'move' to doing a 'copy. This gives the user an additional visual indication that a copy is being performed (other than just the change in the mouse cursor). On drop we either remove this hidden element (for a move) or convert it to a proper item (for a copy).

nrn: Ahh, now I understand! Thanks for explaining.

nrn: Ahh, now I understand! Thanks for explaining.

paul90:
Looking good. Will be nice to get the right visual feedback while copying content.

I wonder if it would be an idea to add the ability of being able to abandon the drag/copy with the  ESC key. This would remove the problem of what to do with an item you are dragging, if you either forget where it was dragged from.

joshuabenuck:
What visual feedback on copy is missing? You should see the mouse cursor indicate that a copy will be performed. The original item should appear or disappear when switching between copy and move.

I’ll see if I can get ESC cancel an in progress drag.

joshuabenuck:
The ESC to cancel drag changes are on http://fed.wiki.randombits.xyz if you want to try them out.

paul90:
> Will be nice to get the right visual feedback while copying content.

I meant when this is merged and released, not that anything is missing.

paul90:
Published as + wiki-client@0.17.0

PR #237, opened 2019-05-03
site adapter promises

nrn:
Paired with @WardCunningham on this.

We are using the returned jquery ajax promises in wiki-plugin-graphviz, this PR makes it unnecessary to pass in a dummy callback, and makes all siteAdapter get methods work with promises the same way the graphviz plugin assumes they do.


paul90:

PR #238, opened 2019-05-12
Restore browser test page

paul90:
For some reason this was over looked when the siteAdapter was created. Here we add mocking of the siteAdapter calls, so that the existing tests run.

### DIFF
--- a/test/neighborhood.coffee
+++ b/test/neighborhood.coffee
@@ -46,7 +46,7 @@ describe 'neighborhood', ->
       expect( searchResult.finds ).to.eql( expectedResult )
 
 
-    it 'searches both the slug and the title'
+    it.skip 'searches both the slug and the title'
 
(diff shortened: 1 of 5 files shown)


WardCunningham:
I am having a little trouble understanding the overall approach here. My presumption would be that a mock would omit network calls for the sake of testing non-network logic. But I see ajax requests in the mock itself? Are these the ajax calls required to get the exist tests to run?

paul90:
Just reusing the server mock that already exists, so provides an easy solution. Would be better to have two versions of the mocked siteAdapter gets, for page found and not found.

WardCunningham:
I'm happy to see this working again. Now I need to build it into my workflow. Better tests are alway an option but fixing this oversight shouldn't wait for that.

paul90:

PR #239, opened 2019-05-12
Allow ESC to save editor.

joshuabenuck:
Request from the group in Dayton to allow ESC to be used to close editors in addition to the current cmd/alt-s key combination.

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -32,6 +32,11 @@ textEditor = ($item, item, option={}) ->
 
   keydownHandler = (e) ->
 
+    if e.which == 27 #esc for save
+      e.preventDefault()
+      $textarea.focusout()
+      return false
+
     if (e.altKey || e.ctlKey || e.metaKey) and e.which == 83 #alt-s for save
       e.preventDefault()
       $textarea.focusout()


joshuabenuck:
The changes have been applied on top of the drag-to-reorder changes on http://fed.wiki.randombits.xyz for those that want to test them.

almereyda:
Thanks for recently contributing the drag-to-reorder function.

Intuitively I would consider ESC to be an "escape" from editing, in so my changes *are not* saved, but actually discarded. A common way to register save actions of text fields that allow line breaks is often also "Ctrl + Return/Enter".

Many journal actions often happen implicitly, e.g. also when clicking outside of a textfield.

I wonder if this proposal could be thought together with attempts at increasing wikis accessibility, in so it also allows to be controlled via the keyboard soley.

WardCunningham:
I google around for guidance and find ESC can mean accept or cancel.

As most computer users are no longer concerned with the details of controlling their computer's peripherals, the task for which the escape sequences were originally designed, the escape key was appropriated by application programmers, most often to mean Stop. [wikipedia](https://en.wikipedia.org/wiki/Esc_key#Uses)

In the insert mode, every character typed is added to the text in the file; pressing the <Esc> (Escape) key turns off the Insert mode. [vi help](https://www.cs.colostate.edu/helpdocs/vi.html)

lib/editor.coffee
```diff
@@ -32,6 +32,11 @@ textEditor = ($item, item, option={}) ->
 
   keydownHandler = (e) ->
 
+    if e.which == 27 #esc for save
+      e.preventDefault()
+      $textarea.focusout()
+      return false
+
     if (e.altKey || e.ctlKey || e.metaKey) and e.which == 83 #alt-s for save
```
paul90: I am minded to change this to be only `e.ctlKey || e.metaKey)` as including alt-s prevent ß from being typed *at least on a UK Apple keyboard*.

paul90: also tempted to remove the other use of alt combination - alt-i and alt-m used for circumflex diacritic and 'µ'.

paul90:

lib/editor.coffee
paul90: This was addressed in https://github.com/fedwiki/wiki-client/pull/240

PR #240, opened 2019-05-15
remove altKey combinations

paul90:
This PR removes the use of the alt key combinations to perform actions on a page item. This allows for these key combinations to revert to their expected behaviour of providing keyboard access to alternative characters.


paul90:

PR #241, opened 2019-05-19
Fix typo in ctrlKey modifier.

joshuabenuck:
After the removal of alt key support in PR #240, I was unable to figure out how to change the type of an item as I used to use alt-m. After looking at the source, I discovered the problem. e.ctlKey is undefined. It needs to be e.ctrlKey.

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -37,18 +37,18 @@ textEditor = ($item, item, option={}) ->
       $textarea.focusout()
       return false
 
-    if (e.ctlKey || e.metaKey) and e.which == 83 #ctrl-s for save
+    if (e.ctrlKey || e.metaKey) and e.which == 83 #ctrl-s for save
       e.preventDefault()
       $textarea.focusout()
       return false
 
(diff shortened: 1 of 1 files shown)


WardCunningham:
This seems to be an old mistake. I've checked, it is e.ctrlKey on my mac also.

nrn:
After some digging, curious whether this was my old type or not, I found the original source: https://github.com/WardCunningham/Smallest-Federated-Wiki/commit/4e89a5bda3ffad4f5710c462871646d2ebc6c824#diff-e6109ed1ee08593abe0e0f971f4bac2bR78

It's been this way for 8 years, and has a note that it didn't work in the beginning 😂Looking through the history we've pretty much all touched/copied/moved this code over the years. The most interesting part though is the comment a few lines above, where at one point it was decided that escape should cancel an edit instead of save, but we didn't implement it.

paul90:
Published as + wiki-client@0.16.1

WardCunningham:
@nrn proves that it was a good idea to save that old repo with all of its history.

PR #243, opened 2019-06-23
expose lineup so plugins can access models

WardCunningham:
We hold page json as js object in two places, `$page.data('data')` and in pageObjects in the lineup. The later is the preferred version but it turns out neither are updated as pages are edited. Here we make the singleton lineup available to plugins so that pageObjects become available and can benefit from anticipated improvements.

With this change the few places that require accurate page json will use 
```
wiki.lineup.atKey($page.data('key'))
```
which will eventually be maintained consistent with the server.

With this change in mind we examine where we might assume that pageObject is already kept current. It turns out not many places but there could be more.

- Legacy: show page source, revision
- Future: some crazy logic that will return to a transporter for more content
- License: some crazy logic that lists authors for CC BY-SA 4.0 link
- PageHandler: fork, pushToLocal (bug)
- Refresh: renderPageIntoElement

So here is the unlikely bug scenario in PageHandler: someone has been successfully editing a page, they lose network connectivity, their next edit causes a pushToLocalStorage, there edits have been lost. Joshua tells me he has experienced this bug while authoring page on the commuter train.

### DIFF
--- a/lib/wiki.coffee
+++ b/lib/wiki.coffee
@@ -171,5 +171,7 @@ wiki.security = require './security'
 
 wiki.createSynopsis = require('./synopsis') ##
 
+# known uses: (none yet)
+wiki.lineup = require './lineup'
 
 module.exports = wiki


nrn:
This looks like a good step forward to me, the more we move over to the lineup model the easier things will be to change in the future.

(though we should try to only commit the package-lock.json file when package.json changes)

paul90:
> (though we should try to only commit the package-lock.json file when package.json changes)

Looks as if that `package-lock.json` commit is changing all the `integrity` fields from `sha-512`, to `sha-1`. Certainly should not be changing to a weaker integrity check.

I see no package version changes in the lock file, so there is nothing to update there, just remove it from the PR.

WardCunningham:
Oops. That package.lock update just happened. It was npm's idea, not mine. Will push a fix and try to keep an eye out for this in the future.

WardCunningham:
I've been reading through old notes and found this mention from many years ago.

http://ward.fed.wiki.org/limit-json-in-the-dom.html

> Used in plugin/reduce to find pageFold of links, 2x.
>
> This last reference is hard to resolve. One approach would be to expose the wiki-client lineup to plugins through the wiki global. This leaves a coordinated release problem between the wiki-client and the plugin-reduce repo.

WardCunningham:
> I expect that we are going to need to co-ordinate updates cross the client and plugins, so do we manage this as a bundle of changes?

My thought was to make this change now and revise plugins later on a time-available basis. I will update the advice offered in plugins.fed.wiki.org.

http://plugins.fed.wiki.org/plugin-runtime-environment.html

Since we are exploring more sweeping changes to the plugin protocols related to async refresh I thought we should get some experience with this kind of coordination. The more serious case is when we change or withdraw features made available to plugins. I would like to make Plugmatic aware of new dependencies on wiki-client and wiki-server versions such that it doesn't offer to install plugins that won't run.

paul90:
That sounds good. 

Let's give this a minor version bump, and maybe then use `peerDependencies` in updated plugins to indicate they need at least a certain version of the client.

WardCunningham:
I can probably get Plugmatic to check peerDependencies and gray out any install that won't work with current installs. Oh, but that probably means I need to fetch the package.json for all versions. Yuck.

paul90:
Thinking about it, I'm not sure that `peerDependencies` provides a solution. The advice in https://nodejs.org/en/blog/npm/peer-dependencies/ is that
> peer dependency requirements, unlike those for regular dependencies, _should be lenient._

**But**, as a package version once published can't be updated, it is not possible to retrospectively possible to modify an over lenient peerDependency. Something that is inevitable going to happen...

nrn:
PeerDependencies is totally the right concept, but it is a very disappointing technical solution :(

PR #244, opened 2019-07-06
Help wiki-server discover static page templates

dobbs:
Partial fix for incompatibility with `pnpm` which Glitch.com uses to install npm packages.

There is a small, related change coming in a sibling PR for `wiki-server`:
https://github.com/fedwiki/wiki-server/pull/147


dobbs:
After some reflection, I think this is a weird dependency for the server to need the client in order to render the static pages.

dobbs:
I think the solution in #245 is better than this one.

PR #245, opened 2019-07-08
Help wiki-server discover static page templates

dobbs:
Enables wiki-server to use `require.resolve('wiki-client')` to find the location of `wiki-client` on disk specifically so express.js can use the `views/static.html` and `views/oops.html` to render pages on the server.

This is an alternative to #244


dobbs:
@nrn shared a trick that makes this PR unnecessary!
https://github.com/fedwiki/wiki-server/pull/148#discussion_r302409596

PR #246, opened 2019-07-22
Prevent duplicate plugin loads using promises

joshuabenuck:
This PR resolves a race condition which was leading to plugins getting loaded multiple times. Promises are a really nice fit here. Eric is going to build on this to add support for plugins that use ES6 modules.

The changes are live on fed.wiki.randombits.xyz if you want to take a look.

FYI: @WardCunningham, @paul90, @dobbs.

### DIFF
--- a/lib/plugin.coffee
+++ b/lib/plugin.coffee
@@ -35,7 +35,7 @@ getScript = plugin.getScript = (url, callback = () ->) ->
         scripts.push url
         callback()
       .fail (_jqXHR, _textStatus, err) ->
-        console.log('Failed to load plugin:', url, err)
+        console.log('getScript: Failed to load:', url, err)
         callback()
 
 plugin.get = plugin.getPlugin = (name, callback) ->
--- a/lib/plugin.coffee
+++ b/lib/plugin.coffee
@@ -12,10 +12,10 @@ escape = (s) ->
     .replace(/'/g, '&#x27;')
     .replace(/\//g,'&#x2F;')
 
-# define cachedScript that allows fetching a cached script.
+# define loadScript that allows fetching a script.
 # see example in http://api.jquery.com/jQuery.getScript/
 
-cachedScript = (url, options) ->
+loadScript = (url, options) ->
   options = $.extend(options or {},
     dataType: "script"
     cache: true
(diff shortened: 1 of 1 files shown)


WardCunningham:
Am I right that the promises created here never leave this code?

The hash loadingScripts holds a promise while the plugin is being loaded and holds the plugin itself after that. A request for a plugin that is found in loadingScripts will wait for the promise to resolve before calling its callback. The key new feature is that multiple callers will wait for a single promise representing a single request for a plugin to be loaded.

joshuabenuck:
@WardCunningham Almost. We do return the promises which enables callers to chain off of the promise themselves without having to pass in a callback. Everything else you wrote is accurate.

From what I can tell, no one ever used the return value from plugin.get so changing it shouldn't break anyone.

lib/plugin.coffee
```diff
@@ -24,24 +24,34 @@ cachedScript = (url, options) ->
...
     callback()
   else
-    cachedScript url
+    loadScript url
       .done ->
         scripts.push url
         callback()
-      .fail ->
+      .fail (_jqXHR, _textStatus, err) ->
+        console.log('Failed to load plugin:', url, err)
```
paul90: Although getScript() is part of the plugin machinery, it is also exposed as wiki.getScript and used elsewhere to load supporting scripts. Might be better to use something like `console.log('getScript: Failed to load:', url, err)`

WardCunningham: Good catch. I'll just make this change to keep this pr moving.

dobbs:
I've deployed a wiki into glitch with this wiki-client code installed
https://great-hat.glitch.me
https://glitch.com/edit/#!/great-hat?path=package.json:12:78
(note for posterity: I change the software on great-hat.glitch.me with abandon, so it's association with this particular PR will be short-lived)

I used https://glitch.com/~wiki-dev to create the custom wiki package that is currently installed at https://great-hat.glitch.me
```
cd ~/wiki-client/
git remote add joshuabenuck https://github.com/joshuabenuck/wiki-client
git fetch joshuabenuck
git checkout -b promisfy-plugin-get --track joshuabenuck/promisfy-plugin-get
npm version prerelease --preid=promisify-plugin-get
npm pack
mv wiki-client-0.18.1-promisify-plugin-get.0.tgz ~/public
cd ~/wiki
git pull
npm install --save https://wiki-dev.glitch.me/wiki-client-0.18.1-promisify-plugin-get.0.tgz
git commit -am "use https://wiki-dev.glitch.me/wiki-client-0.18.1-promisify-plugin-get.0.tgz"
npm version prerelease --preid=promisify-plugin-get
npm pack
mv wiki-0.19.1-promisify-plugin-get.0.tgz ~/public/
```

WardCunningham:
I'm happy to see this merged. I've reviewed the code with Joshua. There is trickier stuff in the pipeline so its a good time to flush out any obscure problems with this straightforward commit before going down that path.

PR #247, opened 2019-08-03
Use custom editor if plugin provides one

dobbs:


dobbs:
Deployed in glich along with an early draft of slackmatic plugin for experimentation.
https://great-hat.glitch.me/view/test-slackmatic

WardCunningham:
Choose slackmatic from the factory menu. See this.
image
Press save. See this.
image
Drop an image. Nothing happens.

I understand that this might be more about enabling a custom editor in factory than about this custom editor. But we should be clear what we would expect of a custom editor and that might be a way to escape from the editor, especially if it has a sequence.

lib/factory.coffee
```diff
@@ -37,11 +37,15 @@ emit = ($item, item) ->
...
         """
     menu.find('a.menu').click (evt)->
-      $item.removeClass('factory').addClass(item.type=evt.target.text.toLowerCase())
+      type = evt.target.text.toLowerCase()
+      $item.removeClass('factory').addClass(item.type=type)
       $item.unbind()
       evt.preventDefault()
       active.set $item.parents(".page")
-      editor.textEditor $item, item
+      wiki.getPlugin(type, (plugin) ->
```
WardCunningham: This async operation introduces a visible delay responding to the factory selection. This is an awkward time to be making this decision.

One possible alternative would be to annotate the factories.json info that is loaded at client-start. The server assembles this file from available plugins.

Along this line, the plugin could offer a page of "Managing Slack Histories" along with the usual About Slackmatic Plugin. If this interface were advertised in factory.json in the `pages:[]` field then the hamburger menu would offer this more descriptive option.

You could convey special behavior in selected instances of Slackmatic by just annotating the item json with extra fields configured by text editing pages/manage-slack-histories.json. Here is an example of this in the (Local) Changes plugin. https://github.com/fedwiki/wiki-plugin-changes/blob/master/pages/submit-changes#L35

dobbs: Yes, this PR is focused on the behavior of factory plugin. The Slackmatic plugin is quite unfinished and provided here as an example of a custom editor.

The factory plugin presents a particular challenge. In order to prevent the visible delay, the custom editor must be loaded by the wiki-client before the author chooses that plugin from the factory. The asynchronous wiki.getPlugin() was the first solution that could possibly work.

I like the suggestion to use factories.json to indicate a custom editor.

I was just looking at how the hamburger menu gets populated:
This code in the client:
https://github.com/fedwiki/wiki-client/blob/master/lib/legacy.coffee#L290-L310
...gets data from this endpoint in the server:
https://github.com/fedwiki/wiki-server/blob/master/lib/server.coffee#L329-L347

I think I'll update this PR to pre-load plugins that declare custom editors in factory.json:
https://github.com/fedwiki/wiki-client/blob/master/lib/legacy.coffee#L20-L21

```
wiki.origin.get 'system/factories.json', (error, data) ->
  window.catalog = data
  window.catalog
    .filter((entry) -> entry.customEditor)
    .forEach(entry -> 
      wiki.getPlugin(entry.name.toLowerCase(), plugin ->
        console.log("#{entry.name} Plugin declares a custom editor, so pre-loading the plugin")
      )
    )
```

lib/legacy.coffee
```diff
@@ -17,8 +17,17 @@ license = require './license'
 asSlug = require('./page').asSlug
 newPage = require('./page').newPage
 
+preLoadCustomEditors = (catalog) ->
+  catalog
+    .filter((entry) -> entry.customEditor)
```
dobbs: as implemented here, the plugin authors include this in `factory.json`
```
  ...
  "customEditor": true,
  ...
```

lib/factory.coffee
```diff
@@ -37,11 +37,17 @@ emit = ($item, item) ->
...
-      $item.removeClass('factory').addClass(item.type=evt.target.text.toLowerCase())
+      pluginName = evt.target.text
+      pluginType = pluginName.toLowerCase()
+      $item.removeClass('factory').addClass(item.type=pluginType)
       $item.unbind()
       evt.preventDefault()
       active.set $item.parents(".page")
-      editor.textEditor $item, item
+      catalogEntry = window.catalog.find((entry) -> pluginName is entry.name)
+      if catalogEntry.customEditor
```
dobbs: here, we consult window.catalog to see if there is a `customEditor`

dobbs:
The updated PR and plugin are now published at glitch to see plugin get pre-loaded
https://great-hat.glitch.me/test-slackmatic.html

lib/legacy.coffee
dobbs: For example, here's slackmatic plugin's factory.json
https://github.com/dobbs/wiki-plugin-slackmatic/blob/master/factory.json

WardCunningham: I like this solution. It offers a nice compromise between built-in plugins and load-on-first-use plugins, the two choices we have today. The server admin is choosing to selectively build-in a capability for all users of that farm only.

```
{
  "name": "Slackmatic",
  "title": "Slackmatic Plugin",
  "category": "other",
  "customEditor": true
}
```
The example I gave was a plugin's about page talking to itself so the choice of terms was of little consequence. Here a plugin is talking to the Factory about Factory issues where everything is custom. Maybe "custom" is redundant or "true" is missing an opportunity to be more precise. I don't have a specific suggestion. Just recognizing the uniqueness of the channel.

WardCunningham:
I may have pontificated in the wrong field to make my thoughts part of our historical record. The curious can consult https://github.com/fedwiki/wiki-client/pull/247#discussion_r310402023.

paul90:
Let's pull that comment into this thread....

> I like this solution. It offers a nice compromise between built-in plugins and load-on-first-use plugins, the two choices we have today. The server admin is choosing to selectively build-in a capability for all users of that farm only.
> 
> ```
> {
>   "name": "Slackmatic",
>   "title": "Slackmatic Plugin",
>   "category": "other",
>   "customEditor": true
> }
> ```
> 
> The example I gave was a plugin's about page talking to itself so the choice of terms was of little consequence. Here a plugin is talking to the Factory about Factory issues where everything is custom. Maybe "custom" is redundant or "true" is missing an opportunity to be more precise. I don't have a specific suggestion. Just recognizing the uniqueness of the channel.

Maybe "custom" is not redundant, it indicates that this parameter is about a custom editor, rather than the standard built-in one. All the core client needs to know is if the plugin supplies a custom editor, so "true" (or "false") is sufficient. As long as there is a standard entry point for the custom editor, that should be quite enough in the core (together with some error handling in case the plugin doesn't really have a custom editor).

dobbs:
I considered naming this configuration `preload` in case plugins should have other reasons to be preloaded. But that seemed premature since the only use case at hand is this editor and I would have needed another configuration for the editor anyway.  Ward has said a couple times that he had expected to see more innovation in editors. Naming this a `customEditor` does promote the existing text editor into The Official Editor where at the moment it's really The Only Editor. Perhaps dropping `custom` from the name is worth it.

WardCunningham:
You are convincing. I'm in favor with no complaints.

dobbs:
With this latest commit, I have simplified the config in factory.json as we discussed here. I've also worked a little harder than usual to provide an informative error message for future plugin authors (like my future self) that might configure the factory but mess up the plugin itself

paul90:
Published as + wiki-client@0.18.1

PR #248, opened 2019-09-01
guard against nulls

WardCunningham:
We can't be sure that page json is well formed. Instead we must make safe assumptions when we find unexpected nulls or undefined. 

```
  wiki-client/lib/resolve.coffee:58
     string = string.replace(/〖(\d+)〗/g, "〖 $1 〗")...
 TypeError: Cannot read property 'replace' of undefined
```

### DIFF
--- a/lib/resolve.coffee
+++ b/lib/resolve.coffee
@@ -9,7 +9,7 @@ module.exports = resolve = {}
 resolve.resolutionContext = []
 
 resolve.escape = escape = (string) ->
-  string
+  (string||'')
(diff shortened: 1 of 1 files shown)


WardCunningham:
I like the discipline of watching for errors and then fixing them. I've made several attempts at schema aware json validators and haven't been happy with any effort yet. This is because the schema ends up more complex than the code reading the json. There are just so many things that can go wrong.

The original wiki was quite careless. If you liked the result it was happy.

I would like to log enough info, site and slug, so we can see the pages that kill the server.

PR #250, opened 2019-11-14
Only cancel drag if escape is pressed.

joshuabenuck:
When code was added to allow item drag operations to be cancelled by pressing escape, it inadvertently broke the ability to copy items between pages.

This PR properly guards the setting of the `dragCancelled` flag so it only happens when escape is pressed.


WardCunningham:
Seems obvious enough and not risky at all. Let's merge.

paul90:
Looks good. Merged.

PR #251, opened 2019-11-15
Prevent page from being dropped on a journal

joshuabenuck:
When reordering pages, it is sometimes possible to get the page to drop onto the journal of an adjacent page. The easiest way to reproduce this is to scroll down so a large journal is visible and then drag another page across the page with the large journal. The large journal will get a small outline when it is the drop target. If you drop the page at this point, the page freezes (example screenshot below).

image

This is the error that appear in the JavaScript console.
```
client.max.js:2714 Uncaught TypeError: Cannot read property 'getRawPage' of undefined
    at Object.merge (client.max.js:2714)
    at HTMLDivElement.j (client.max.js:3630)
    at e.<computed>.<computed>._trigger (jquery-ui.min.js:6)
    at e.<computed>.<computed>._drop (jquery-ui.min.js:7)
    at e.<computed>.<computed>._drop (jquery-ui.min.js:6)
    at e.<computed>.<computed>.<anonymous> (jquery-ui.min.js:7)
    at Function.each (jquery-2.2.4.min.js:2)
    at Object.drop (jquery-ui.min.js:7)
    at e.<computed>.<computed>._mouseStop (jquery-ui.min.js:8)
    at e.<computed>.<computed>._mouseStop (jquery-ui.min.js:6)
```

This is prevented by configuring the journal's droppable to only accept other journals (this is done using the `accept` property of the droppable).
https://api.jqueryui.com/droppable/#option-accept

Another change is the addition of a style to constrain the width of the dragged journal so it doesn't flatten to a single row. This both looks nicer and is easier to manipulate. Obligatory screenshoots showing the effects of the style change below.

Before:
image

After (ignore the yellow highlights - artifact of keys pressed to trigger the screenshot):
image

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -162,6 +162,10 @@ footer {
   padding: 1px;
 }
 
+.journal.ui-draggable {
+  width: 420px;
+}
+
 .action.fork {
   color: black; }
 
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -119,7 +119,7 @@ $ ->
     return false
 
   $('.main')
-    .sortable({handle: '.page-handle', cursor: 'grabbing'})
+    .sortable({handle: '.page-handle', cursor: 'grabbing', cancel: 'journal'})
       .on 'sortstart', (evt, ui) ->
         return if not ui.item.hasClass('page')
         noScroll = true
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -119,7 +119,7 @@ $ ->
     return false
 
   $('.main')
-    .sortable({handle: '.page-handle', cursor: 'grabbing', cancel: 'journal'})
+    .sortable({handle: '.page-handle', cursor: 'grabbing'})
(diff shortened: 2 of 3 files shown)


WardCunningham:
Two very nice catches.

The first case was never considered. This being a rarely or never used feature, no one complained.

The second case must be rooted in slowly evolving interpretation of css/js/jquery. As coded at a hackfest it looked right and demo'd well. As years went by this behavior changed, probably due to a more strict interpretation of something or other. Thank you for the addition that returns the intended presentation.

Aside: Once you see the journal drag thing work once or twice then you too will want it to work every time. Experience shows this to not be a key feature like git merge, but it could be something we learn to include in many workflows.

WardCunningham:
I haven't downloaded and tested this pull request but I have inspected the code and find it to be suitable for merge as submitted. The commits complete api calls and extend very focused declarations making for very low risk even if they didn't work exactly as intended.

paul90:
Looks good. Merged.

PR #252, opened 2019-11-18
Highlight items when their bind is blocked

joshuabenuck:
This PR reduces the requirements on plugin authors while also making the system easier to debug.

Prior to these changes, plugin authors had to remember to put a `consuming` property on their items as part of their bind. This was a manual, error prone process.

After this change, the bind wrapper in plugin takes care of this for them, but there is a slight difference in behavior.

The old behavior only tagged an item as consumed after it was consumed. The new behavior tags an item as consumed the moment something else is waiting on it. Without this change, it isn't possible to highlight what is blocking the bind of the items highlighted in red. With this change, shift hovering over an item in red will highlight in orange the items being consumed. At least one of those highlighted in orange will be the thing blocking the bind.

Here's a screenshot taken by reintroducing the lack of resolve on error defect in the datalog plugin.
image

Shift hovering over either of the tail items will highlight all datalogs that are up and to the left of the hovered tail.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -249,6 +249,9 @@ footer {
 .consumed {
   outline: solid 2px orange !important; }
 
+.highlight {
+  outline: solid 2px red; }
+
 .report p,
 .factory p,
 .data p,
--- a/lib/plugin.coffee
+++ b/lib/plugin.coffee
@@ -90,6 +90,7 @@ bind = (name, pluginBind) ->
     # Wait for all items in the lineup that produce what we consume
     # before calling our bind method.
     if consumes
+      $item[0].consuming = []
       deps = []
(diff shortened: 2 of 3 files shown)


joshuabenuck:
It is worth noting that plugin authors can redefine the list of consumed items in their bind. They just have to be sure to clear the list set by the bind wrapper. This allows them to indicate they are only consuming a subset of the applicable items in the lineup. The `tail` plugin needs to be slightly updated based on this PR as it relies on consuming only containing the item actively being tailed.

joshuabenuck:
Update to the tail plugin that takes this PR into account for those interested: https://github.com/joshuabenuck/wiki-plugin-tail/commit/84a399c0133ee2e7ae5e82aabd8e0fea1a747693

PR #253, opened 2019-11-19
Fix pressing enter to create new editors and clean up event handlers

joshuabenuck:
When pressing enter in an editor, a new editor usually opens. With the most recent changes, this breaks.

This PR removes what appears to be an unnecessary call to `plugin.do`. That call is asynchronous and rerenders the lineup which stomps on the editor that is created right after the call.

It also calls `.off()` before rendering an item. This cleans up any event handlers previously attached to the item itself.

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -142,7 +142,6 @@ spawnEditor = ($page, $before, type, text) ->
     .data('item', item)
     .data('pageElement', $page)
   $before.after $item
-  plugin.do $item, item
   before = itemz.getItem $before
   textEditor $item, item, {after: before?.id}
 
--- a/lib/plugin.coffee
+++ b/lib/plugin.coffee
@@ -51,6 +51,7 @@ plugin.renderFrom = (notifIndex) ->
     item = $item.data('item')
     promise = promise.then ->
       return new Promise (resolve, reject) ->
+        $item.off()
         plugin.emit $item.empty(), item, () ->
           resolve()
     emitNextItem(itemElems)


WardCunningham:
The text editor has two features, split and join, that only apply to Paragraph and more recently to Markdown. This makes four cases that should be tested to be sure they write the right thing into the Story, the Journal, and the server.

Now, from memory, Paragraph splits on return with the cursor placed anywhere and joins on backspace with the cursor at the first position and only when joining another Paragraph.

Paragraph is unique in resisting newlines as markup.

Markdown accepts newlines on input but interprets Return Return as split. I can't remember all the rules that make this work but it is trying to be like split and join for Paragraph.

joshuabenuck:
Note: The line I removed has been there for years.

This will likely prompt the question of why it hasn't been a problem before now.

Prior to the bind ordering work, calling `plugin.do` on something unnecessarily didn't have any lasting consequences as it only impacted a single item. The new behavior of `plugin.do` causes it to rerender the lineup starting at the target item and it does so asynchronously.

My first attempt at fixing this resulted in a duplicate item being added to the lineup as what we passed to `plugin.do` was a temporary item that we then immediately passed to the text editor.

I have been unable to think of a use case that would benefit from rendering an item only to immediately edit the same item. Should one exist, this change will break it.

Thanks for the use cases @WardCunningham. That will help in the search for other things that could be impacted by this change.

WardCunningham:
Well, there is the tricky way that the Factory menu becomes some other Item type when something is selected. Also the way Cmd-M returns to the Factory to select an alternate item type. I seem to recall that you have improved the logic in this area so I doubt I am telling you anything new. Still, many intricate cases to be tested when working in this area.

joshuabenuck:
*Good news:*
Editing paragraph and markdown items appear to function normally. The entries in the journal look correct. Cmd-M also functions correctly. All changes look to be persisted properly server side.

*Bad news:*
The plugin bind ordering re-renders items down and to the right of any changed item. This calls emit and bind for those items, but does not clear out the event handlers. Prior to bind ordering, this was never an issue because we only ever re-rendered when we edited an item. That process calls `.empty()` on the item which takes care of cleaning up the event handlers.

This PR is good as is. I'll submit a separate PR once I figure out how to clear out the event handlers.

joshuabenuck:
The fix to clean up the event handlers is only a single line. I've added it to this PR since it is a small change and is loosely related to getting editors to work properly. The title and initial description have been updated to reflect the expanded scope.

My previous comment originally made the claim that the reason editing an item cleaned up its event handlers was because we called `.empty()` on the item before editing it. That was incorrect. `.empty()` is called both when editing and rerendering the lineup.

`.empty()` doesn't clean up event handlers on the item itself, only its children. To clean up the handlers on the item, `.off` needs to be called. I have updated my previous comment to reflect what actually happens.

Here are the relevant excerpts from the jquery api docs.

https://api.jquery.com/empty/
> .empty()
> Description: Remove all child nodes of the set of matched elements from the DOM.
> To avoid memory leaks, jQuery removes other constructs such as data and event handlers from the child elements before removing the elements themselves.

https://api.jquery.com/off/
> .off( events [, selector ] [, handler ] )
> The .off() method removes event handlers that were attached with .on().
> Calling .off() with no arguments removes all handlers attached to the elements.

texteditor.coffee calls `.unbind()` instead of `.off()`. This PR uses `.off()` because `.unbind()` is deprecated.

https://api.jquery.com/unbind/
> As of jQuery 3.0, .unbind() has been deprecated. It was superseded by the .off() method since jQuery 1.7, so its use was already discouraged.

PR #255, opened 2019-11-30
prevent promise return when callback provided

WardCunningham:
This prevents an unhandled promise error message in the log. From #237.

image

Mob programmed with Eric, Joshua and Paul.


WardCunningham:
I was reviewing this pull request with Thompson and noticed that our second two modifications were omitted. So we must give him credit as a reviewer too.

PR #258, opened 2019-12-07
Initial implementation of incremental search

joshuabenuck:
Putting this up for review. Typing in the search box immediately runs a search. If the search returns multiple results, pressing enter will open a search results page. If there's only one, the single page in the results will be opened instead.

Work remaining:
* Tighten up a couple of CSS queries.
* Possibly tweak the styling.
* Perform more testing (need to test with search boxes on other pages open).

image

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -70,6 +70,10 @@ footer form, .footer-item {
   font-style: italic;
 }
 
+.omitted-results {
+  font-style: italic;
+}
+
 .incremental-search {
   position: absolute;
   background-color: white;
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -80,6 +80,7 @@ footer form, .footer-item {
   color: black;
   display: inline-block;
   max-height: 300px;
+  max-width: 430px;
   padding-top: 5px;
   padding-bottom: 5px;
   padding-right: 10px;
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -66,6 +66,23 @@ footer form, .footer-item {
 
 .footer-item { margin-left: 8px; }
 
+.incremental-search {
+  position: absolute;
+  background-color: white;
+  color: black;
+  display: inline-block;
+  max-height: 300px;
+  padding-top: 5px;
+  padding-right: 10px;
+  padding-left: 10px;
+  box-shadow: 0px 0px 2px rgba(0, 0, 0, 0.4);
+  outline: none;
+}
+
+.search-term {
+  background-color: yellow;
+}
+
 .neighbor {
   float: right;
   padding-left: 8px;
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -66,6 +66,10 @@ footer form, .footer-item {
 
 .footer-item { margin-left: 8px; }
 
+.no-results {
+  font-style: italic;
+}
+
 .incremental-search {
   position: absolute;
   background-color: white;
@@ -73,10 +77,13 @@ footer form, .footer-item {
   display: inline-block;
   max-height: 300px;
   padding-top: 5px;
+  padding-bottom: 5px;
   padding-right: 10px;
   padding-left: 10px;
   box-shadow: 0px 0px 2px rgba(0, 0, 0, 0.4);
   outline: none;
+  overflow-y: auto;
+  overflow-x: hidden;
 }
 
 .search-term {
--- a/lib/search.coffee
+++ b/lib/search.coffee
@@ -4,11 +4,32 @@
 
 pageHandler = require './pageHandler'
 random = require './random'
-plugin = require './plugin'
 link = require './link'
 active = require './active'
 newPage = require('./page').newPage
+resolve = require './resolve'
+page = require './page'
 
+# From reference.coffee
+emit = ($item, item) ->
+  slug = item.slug
+  slug ||= page.asSlug item.title if item.title?
+  slug ||= 'welcome-visitors'
+  site = item.site
+  resolve.resolveFrom site, ->
+    $item.append """
+      <p>
+        <img class='remote'
+          src='#{wiki.site(site).flag()}'
+          title='#{site}'
+          data-site="#{site}"
+          data-slug="#{slug}"
+        >
+        #{resolve.resolveLinks "[[#{item.title or slug}]]"}
+        —
+        #{resolve.resolveLinks(item.text)}
+      </p>
+    """
 finishClick = (e, name) ->
   e.preventDefault()
   page = $(e.target).parents('.page') unless e.shiftKey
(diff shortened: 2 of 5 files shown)


paul90:
Getting some strange, and inconsistent, behaviour with this...

Text from the search results spilling out results box, this can result in when search box looses focus it is not possible to get back to the search box.
Screenshot 2019-12-09 at 10 22 50

Sometimes clicking on a result will refresh the page with just that wiki page, rather than adding it to the lineup. Not managed to get this happen consistently. Can happen when clicking on a result, and also on any link in the displayed synopsis.

There should be a way of dismissing the search results, other than having to clear the search box. Elsewhere pressing `ESC` or the search box loosing focus appear to be the standard.

client/style/style.css
```diff
@@ -66,6 +66,23 @@ footer form, .footer-item {
...
+  position: absolute;
+  background-color: white;
+  color: black;
+  display: inline-block;
+  max-height: 300px;
+  padding-top: 5px;
+  padding-right: 10px;
+  padding-left: 10px;
+  box-shadow: 0px 0px 2px rgba(0, 0, 0, 0.4);
+  outline: none;
```
paul90: adding `overflow: hidden;` should help with text spilling out of the search results.

joshuabenuck: I added `overflow-y: auto` (to allow a scrollbar to be shown if and only if it is needed) and `overflow-x: hidden`.

joshuabenuck:
Here's what is displayed when a search returns no results.
image

joshuabenuck:
> There should be a way of dismissing the search results, other than having to clear the search box. Elsewhere pressing `ESC` or the search box loosing focus appear to be the standard.

I completely agree. `ESC` and clicking on a page in the lineup will now close the incremental search results. (I tried adding closing when the browser window loses focus, but that felt too aggressive so I removed it)

joshuabenuck:
> There should be a way of dismissing the search results, other than having to clear the search box. Elsewhere pressing `ESC` or the search box loosing focus appear to be the standard.

I completely agree. `ESC` and clicking on a page in the lineup will now close the incremental search results. (I tried adding closing when the browser window loses focus, but that felt too aggressive so I removed it)

joshuabenuck:
Number of displayed results limited to **100**. This is what is displayed when results are omitted.
image

joshuabenuck:
> Sometimes clicking on a result will refresh the page with just that wiki page, rather than adding it to the lineup. Not managed to get this happen consistently. Can happen when clicking on a result, and also on any link in the displayed synopsis.

I had to add click handlers to get pages to open in the lineup. I wonder if there's an edge case that isn't handled by the code I pulled in from `legacy.coffee`.

paul90:
Think I may have managed to reproduce the issue with opening a page, rather an added the wiki page to the line-up. When it happens I am seeing
Screenshot 2019-12-10 at 11 33 27

This looks to be here.. https://github.com/fedwiki/wiki-client/blob/b1c984b163bf0f0de3a10ee8c475a3819a20adf7/lib/search.coffee#L33

Have no explanation why it happens, as it sometimes works correctly, and sometimes doesn't...

joshuabenuck:
@paul90 Thanks! That gave me what I needed to track down the problem. It occurs when you click on the highlighted portion of the search result and not when clicking on the unhighlighted portion.

My latest checkin fixes that and two other problems:
* Double clicking on a search result would open the text editor on that result.
* The case of the search term in the synopsis matched the case of the term in the search instead of in the result.

lib/search.coffee
```diff
@@ -72,9 +73,14 @@ createSearch = ({neighborhood})->
...
               else return p
             .join('')
-          text: result.page.synopsis.split(new RegExp(searchQuery, 'i')).join("{{#{searchQuery}}}") || '',
+          text: result.page.synopsis
+            .split(new RegExp("(#{searchQuery})", 'i'))
+            .map (p) ->
+              if searchQuery.toLowerCase() == p.toLowerCase()
+                return "{{#{p}}}"
+              else return p
+            .join('')
```
WardCunningham: I'm trying to follow along at home. Can you remind me what {{ }} does in this code? How does this come to have the undesirable effect you corrected?

joshuabenuck: `{{ }}` are markers that surround the search term so I can find it after it has gone through rendering by the `reference` plugin (without the risk of breaking the markup generated by the `reference` plugin). This particular `if` clause was added to preserve the case of the search term as it appears in the original source rather than using the case of the search term as typed by the user.

The current version of this code does the following:
* Perform a capturing case insensitive match of the search term.
* Surround all instances of the match with the `{{ }}` markers while preserving the casing of the search term from the original source.
* Render the search matches through the `reference` plugin.
* Find all instances of the now rendered search terms and surround them with a `<span/>` to add highlighting.

lib/search.coffee
```diff
@@ -2,15 +2,102 @@
...
+        p.emit($item, item)
+        $item.html($item.html()
+          .split new RegExp("(\{\{#{searchQuery}\}\})", 'i')
```
joshuabenuck: There's a flaw here. This needs to be a `.*` and not `searchQuery`. If the rendering by the `reference` plugin adds any markup to the search term, this will not find those matches.

lib/search.coffee
WardCunningham: Ah. I knew there was something to it that I didn't yet appreciate. Thanks for explaining.

WardCunningham:
It might make sense to delay the incremental search until at least a couple of characters have been typed. I insist on at least two characters in my reimplementation of title-search for the original wiki.

http://wiki.c2.com/?FindPage

https://github.com/WardCunningham/remodeling/blob/c95e80ebfe685aa7b307916c2127cee8f5ebc593/docs/index.html#L168-L173

joshuabenuck:
@WardCunningham Search is now delayed until at least two characters have been typed. You can try it out on my [dev wiki](https://dev.wiki.randombits.xyz/view/welcome-visitors#). There's a roster on the welcome page so you can easily populate the neighborhood and get more to search.

While my approach of using `{{ }}` to avoid munging markup seems to work, it doesn't handle searching for phrases that span links. Here's an example.

Searching for `this` works (or even searching for `Federated Wiki`).
image

But searching for `this Federated` does not.
image

I bring it up to gauge how others feel about the limitations of the current approach. I see no simple fix. All of the approaches I can think of involve in depth parsing of either the source markup or the rendered HTML. It is also worth adding that the current behavior looks to match that of the exiting search.

WardCunningham:
We should make the best use possible of information in the sitemaps and if this disappoints (unlikely) then consider improvements there.

paul90:
Published together with #259 as + wiki-client@0.20.0

PR #259, opened 2020-01-18
Full text search

paul90:
This builds on #258, working together with fedwiki/wiki-server#152 to add full text search for the neighbourhood. Here we are using [minisearch](https://github.com/lucaong/minisearch).

For full text search we rely on the server constructing an index and serializing it, so we can download it and load it in the browser. Where this index is not available we use the synopsis from the sitemap to construct an index for the wiki.

We then use the code from #258 to do an incremental search across this indexes, the results are sorted by rank for presentation.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -66,6 +66,14 @@ footer form, .footer-item {
 
 .footer-item { margin-left: 8px; }
 
+.no-results {
+  font-style: italic;
+}
+
+.omitted-results {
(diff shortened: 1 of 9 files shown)


paul90:
It is worth noting that the commits prior to d8b4511 are part of #258.

paul90:
Published together with #258 as + wiki-client@0.20.0

PR #262, opened 2020-02-23
decode urls based on image sufix

WardCunningham:
Work in Progress

This should be enough to drag image urls to a factory and get an Image item that cites the remote resource. We've resisted this before because the remote image can disappear. This is a step forward toward more advanced image handling.

Aside: I couldn't demonstrate even this much in my development environment. I'd love to pair in order to get past this inability.

Some further work we'd like to explore:

- [x] fetch the image in some cors compatible way and validate that it is an image.
- [x] fetch the image and store a thumbnail in the image that won't go away
- [ ] fetch the image and save it in the image assets folder
- [x] bypass cors limitation by enriching Image items server side

### DIFF
--- a/lib/drop.coffee
+++ b/lib/drop.coffee
@@ -24,6 +24,12 @@ isPage = (url) ->
     return item
   null
 
+isImage = (url) ->
+  parsedURL = nurl.parse(url, true, true)
+  if parsedURL.pathname.match(/\.(jpg|jpeg|png|svg)$/)
+    return url
+  null
+
 isVideo = (url) ->
   parsedURL = nurl.parse(url, true, true)
   # check if video dragged from search (Google)
@@ -68,6 +74,9 @@ dispatch = (handlers) ->
       if video = isVideo url
         if (handle = handlers.video)?
           return stop handle video
+      if image = isImage url
+        if (handle = handlers.image)?
+          return stop handle image
       punt = {url}
     if file = isFile event
       if (handle = handlers.file)?
--- a/lib/drop.coffee
+++ b/lib/drop.coffee
@@ -26,7 +26,7 @@ isPage = (url) ->
 
 isImage = (url) ->
   parsedURL = nurl.parse(url, true, true)
-  if parsedURL.pathname.match(/\.(jpg|jpeg|png|svg)$/)
+  if parsedURL.pathname.match(/\.(jpg|jpeg|png|svg)$/i)
     return url
   null
 
--- a/lib/factory.coffee
+++ b/lib/factory.coffee
@@ -229,8 +229,8 @@ resizeImage = (dataURL) ->
   cW = undefined
   cH = undefined
   # target size
-  tW = 800
-  tH = 512
+  tW = 500
+  tH = 300
   # image quality
   imageQuality = 0.5
 
--- a/lib/factory.coffee
+++ b/lib/factory.coffee
@@ -108,9 +108,9 @@ bind = ($item, item) ->
     fetchRemoteImage(url)
       .then (dataURL) ->
         resizeImage dataURL
-      .then (resizedURL) ->
+      .then (resizedImageURL) ->
         item.type = 'image'
-        item.url = resizedURL
+        item.url = resizedImageURL
         item.source = url
         item.caption ||= "Remote image"
         syncEditAction()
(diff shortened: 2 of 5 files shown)


paul90:
I have had a few improvements to how images are handled floating around in my thoughts for a while, in addition to those already mentioned...

**Custom Editor:** Use the feature added in #247 to have a image plugin editor to complement the use of a DSL.

**Journal:** This has wider implications, but it would be good to only include the plugin attributes that have changed. 

Having a sparse journal makes having the image correctly sized for the space it is going to occupy somewhat less important, but only having thumbnail sized image stored on the wiki page with any other sizes/original stored as an asset will likely help keep page size manageable.

**Image Placement:** Currently the image plugin placement is somewhat less than idea, only supporting images floating on the right. This results in, when two images placed next to each other, the one on the left appearing after the one on the right when looking at the items in the story. Any other placement requires the user to use the html plugin. It would be good add support for both wiki page width images, and better side by side placement.

paul90:
Promise based `resizeImage`, the promises resolve when image `onload` events have a completed.

Not yet looked at making the initial squeeze to a multiple of the desired final image size.

WardCunningham:
I can understand that network fetch of bits must be async but I am surprised that each squeeze iteration must be async also. Is this because some dom operation is similarly async?

paul90:
Yes, `image.src` load is async - with the `onload` event firing once the image has loaded. So, the `onload` get fired when the freshly squeezed image has been loaded.

WardCunningham:
How did you ever figure this out?

How did this code ever work?
```
  img = new Image
  img.src = dataURL
  return dataURL if smallEnough img
```

paul90:
Added some to set the size on the first squeeze, so we will get an image with the target width, or height - rather than something smaller.

I think this is ready to go for adding image url drop, and as a fix for the squeeze problem.

WardCunningham:
I see you have increased the target dimensions to 800x512 from our previous 500x300. I consider this generous considering that our previous squeeze might end up at halve of those dimensions. We should tune these dimensions and the jpeg quality such that ultimate image size land around 20k to 40k before url encoding. This will change, of course, once we engage by retrieving the source image.

paul90:
The size got changed when looking at getting a better image for using in the dialog. Was also part of experimenting in getting a good size for the initial squeeze.

WardCunningham:
Let's change the image plugin to use the source url if present. This would go right next to, or in place of, this ipfs logic. All of this is in wiki-client so goes together nicely. We can get some experience with large images in assets by dragging them there first and then dragging the exposed link to a Factory.

https://github.com/fedwiki/wiki-client/blob/26cbedcd7764e55be44ebc3f1d9beb0943d457db/lib/image.coffee#L26
The probe for an ipfs server clutters up everyone's console log with error messages that can't be ignored.

WardCunningham:
I set out to add the logic to image.coffee suggested above. I started by testing the branch as it now stands and am unsure that what I am seeing is as intended.

My test case is this  546 × 480 pixels image from commons. (Excuse the vanity.)
https://upload.wikimedia.org/wikipedia/commons/thumb/a/a5/Ward_Cunningham_in_San_Francisco.jpg/546px-Ward_Cunningham_in_San_Francisco.jpg

I expected it to squeeze the image to some power of two multiple of the target size 500 x 300. Since this image is already close I expected the first squeeze to be 500 x something, or, something by 300 which would be small enough that no subsequent 2x squeezing would be required.

I find instead an even smaller image as the final result, 176 x 156.

```
$('.image:last img').get(0).width
176
$('.image:last img').get(0).height
156
```
Having read through the code I now see that there are many additional considerations such as "natural" dimensions. How confusing. I've added many console.log but am still having trouble understanding the numbers. I'll push my debug code and return to this when fresh in the morning.

paul90:
That is most likely due to altering the target size, without revisiting the size for the first squeeze. The width and height for the first squeeze needs to be such that with repeated halving that one dimension will exactly match its target before the other is smaller than its target.

WardCunningham:
Let's merge and publish this version. We can go for smaller thumbnails once we know the source has been captured and will be consulted broadly in the federation.

paul90:
Published as + wiki-client@0.20.2

cliveb:
Thank you for the image squeeze fix. I get carried away writing sometimes inserting images ( http://clive.tries.fed.wiki/view/two-phase-commit-vs-polymerase-function )

PR #263, opened 2020-03-21
extend site adapter to add getIndex to get the serialized site index.

paul90:
This extends the site adapter, to add `getIndex` to enable the serialized site index file to be fetched as text rather than converting it to javascript object.

This fixes an issue seen with sites that require login to view.

### DIFF
--- a/lib/neighborhood.coffee
+++ b/lib/neighborhood.coffee
@@ -34,17 +34,14 @@ populateSiteInfoFor = (site,neighborInfo)->
         wiki.site(site).refresh () ->
           # empty function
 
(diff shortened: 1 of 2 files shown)


paul90:
Published as + wiki-client@0.20.1

WardCunningham:
Installed and confirmed to work for ward.nrrd.wiki.org.

Console log when expanding roster:
```
nrrd.wiki.org index loaded
eric.nrrd.wiki.org index loaded
jeff.nrrd.wiki.org index loaded
joshua.nrrd.wiki.org index loaded
beth.nrrd.wiki.org index loaded
```

Neighborhood pages counts non-zero.

Full test search finds all neighbors.

PR #264, opened 2020-03-24
Highlight search terms in results

paul90:
Highlight the terms used in the progressive search, rather than just matching phrases.

Screenshot 2020-03-24 at 09 29 32


paul90:
Published as part of + wiki-client@0.20.2

Issue #266, opened 2020-06-11
Uneditable paragraphs


ig3:
Sometimes I am unable to edit a paragraph. When I double-click on the paragraph, nothing obvious happens. In particular, the paragraph is not replaced with a textarea containing its content.

When this happens, it is a persistent state. I can edit other paragraphs on the page but when I come back to the uneditable paragraph, it remains uneditable.

After closing and re-opening the page I am able to edit the paragraph again.

Sometimes several paragraphs on a page will be uneditable.

I don't know how to reliably reporduce the problem. It doesn't happen very often.

ig3:
I have just discovered a way to, fairly reliably, reproduce the problem:

1) double-click on a paragraph to edit it
2) double-click on a preceding or following editable paragraph to edit it, while the textarea of the first paragraph remains open

After this, double-click on the first paragraph, whether or not any other paragraph is being edited at the time, has no obvious effect.

WardCunningham:
Each plugin is responsible for handling events and will typically register interest in double-click as they emit and bind their representation in the dom. This happens when a page is viewed and after each markup edit. It is possible that the registration has been lost. Perhaps a javascript error of some type prevented the emit and bind from completing.

Is it always the same plugin? Does it happen with different browsers? Do you see errors in the javascript console?

WardCunningham:
I tried to reproduce the problem with three small paragraphs on a new page using your method. I don't see the problem with an older release of wiki but do see it with a newer release.

WardCunningham:
This is my test case mentioned above.  My test is to refresh the page, double-click the first paragraph, then double-click the second paragraph. I expect the second double-click to close the first editor on loss of focus and then open the the editor on the second paragraph. In both cases I report the javascript console output while clicking. 

This site works correctly.
http://found.ward.bay.wiki.org/view/test
```
textEditor 828a01ea63f4812f {append: true}
textEditor 947c01384a4e6f6e {append: true}
```

This site fails by the second double-click.
http://ward.asia.wiki.org/view/test
```
textEditor 828a01ea63f4812f {append: true}
textEditor 947c01384a4e6f6e {append: true}
notifIndex 0 about to render (3) [div.item.paragraph.ui-sortable-handle, div.item.paragraph.ui-sortable-handle.textEditing, div.item.paragraph.ui-sortable-handle]
```

[...]

PR #267, opened 2020-06-14
skip item being edited in renderFrom()

paul90:
Looks to be a fix for #266.

### DIFF
--- a/lib/plugin.coffee
+++ b/lib/plugin.coffee
@@ -48,12 +48,13 @@ plugin.renderFrom = (notifIndex) ->
     return promise if itemElems.length == 0
     itemElem = itemElems.shift()
     $item = $(itemElem)
-    item = $item.data('item')
-    promise = promise.then ->
-      return new Promise (resolve, reject) ->
-        $item.off()
-        plugin.emit $item.empty(), item, () ->
-          resolve()
+    unless $item.hasClass('textEditing')
+      item = $item.data('item')
+      promise = promise.then ->
+        return new Promise (resolve, reject) ->
+          $item.off()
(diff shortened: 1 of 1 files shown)


ig3:
This does fix #266 for me.

paul90:
As far as I can see it will. Just waiting for it to be double checked

WardCunningham:
I'm pleased to have the chance to apply the neat feature of github where it ignores whitespace changes and shows this to be the single line change that it is.

https://github.com/fedwiki/wiki-client/pull/267/commits/8d8dbd83570115a77c583082b2a1b5db742f7ec4?diff=unified&w=1

This seems safe in that it handles a case that I doubt what considered when this rendering loop was coded. I'm happy to see this merged knowing that any future developer would need to "blame" the history to find the unexpected consequences of following the now excluded statements.

Let's give @joshuabenuck a day or two to look at this because he is more aware of the many cases in which a page is rendered. We overlap some activity with the server turnaround that writes an action and that means we are expecting (maybe depending upon) relative quiescence while this happens.

joshuabenuck:
My first thought of a fix was to put a guard around plugin.do in [editor.coffee](https://github.com/fedwiki/wiki-client/blob/master/lib/editor.coffee#L103), but it isn't immediately obvious what such a guard would look like.

The approach used here is good. I think it is simpler than the one I was considering above. Once the edit is finished, the item that was skipped will be rerendered when the focusout handler is called so there shouldn't be any downsides.

paul90:
Thanks for looking at that @joshuabenuck.

paul90:
Published as + wiki-client@0.20.3

PR #270, opened 2020-12-31
Add Backlinks and resolve a siteAdapter issue

paul90:
This PR adds backlinks to a wiki page. This relies on extra link information being in the sitemap, added by the changes in fed-wiki/wiki-server#161
Backlinks are added at the bottom of the wiki page, just above the journal if it is being shown, initially in a collapsed state:
Screenshot 2020-12-31 at 17 18 44
This expands to show a collaborative link for the page linking, and flags for the sites the page with links was found on:
Screenshot 2020-12-31 at 17 18 58
Clicking on the collaborative link in the list behaves exactly as if the link was within the page. Click on the flag will open page from the wiki that flag is for. 

Holding shift and hovering over the flag will highlight, and try to align, the page item containing the link. Just like it does with page items and journal items already.

A bug was discovered in the siteAdapter, where if there was an error saving how to access a remote site it would incorrectly set the site inaccessible. This appears to happen when the adapter has been called more that once for a site. This has addressed by using the value we have, and is stored in memory, and just reporting that the issue with the persistent store.

Also in the siteAdapter, some promise returns are added that appear to have been missed when others were added a while ago.

The changes in this PR are intended for use along with the changes in fedwiki/wiki-server#161

### DIFF
--- a/client/style/print.css
+++ b/client/style/print.css
@@ -47,6 +47,10 @@ A:link, A:visited {
 	font-weight: 500;
 }
 
+.backlinks {
+	display: none;
+}
+
 .journal {
 	display: none;
 }
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -191,7 +191,7 @@ footer {
   font-size: 80%;
 }
 
-.backlinks h3 {
+.backlinks summary {
   margin-top: 5px;
   margin-left: 5px;
   margin-bottom: 3px;
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -181,10 +181,35 @@ footer {
 .image .thumbnail {
   width: 100%; }
 
+.backlinks {
+  width: 420px;
+  margin-top: 2px;
+  clear: both;
+  background-color: #eeeeee;
+  padding: 1px;
+  color: grey;
+  font-size: 80%;
+}
+
+.backlinks h3 {
+  margin-top: 5px;
+  margin-left: 5px;
+  margin-bottom: 3px;
+}
+
+.backlinks div {
+  margin-left: 5px;
+  margin-right: 2px;
+}
+
+.backlinks a {
+  color: steelblue;
+}
+
 .journal {
   /*width: 420px;*/
   overflow-x: hidden;
-  margin-top: 2px;
+/*  margin-top: 2px; */
   clear: both;
   background-color: #eeeeee;
   overflow: auto;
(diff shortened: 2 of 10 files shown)


WardCunningham:
Some things to consider but not problems that would block a merge.
- [ ] does the ui handle the case when there are more sites than fit? Example would be who links to How to Wiki? Every site with unmodified Welcome Visitors has this link. Recent Changes puts a cap on how many flags are shown. I think it even favors flags earlier in the neighborhood since they might be more important.
- [ ] can it be made clear that sites without this feature won't be found? For example, the neighborhood could show flags smaller for sites that don't support this feature. A better idea would be welcome.
- [ ] item ids can be aliased so that no id is duplicated. There is some logic by which ids are restored when the need for aliasing has passed. A field 'aka' saves the permanent id.
- [ ] also concerns for adding item id to sitemaps mentioned in companion wiki-server pull request.

WardCunningham (approved):

paul90:
> does the ui handle the case when there are more sites than fit?

Good question. Would make sense to handle the same was as in the Activity Plugin, which just takes the first 10. I'll look at getting that included.

> can it be made clear that sites without this feature won't be found?

Not sure. A change in the flag size would give a visual clue, and not be intrusive. Maybe adding some text in the alt text to indicate that link details are not available from that wiki. Will have a think.

WardCunningham:
> Maybe adding some text in the alt text to indicate that link details are not available

This is a good suggestion. I propose positive language, something extra in the description. Brief and unambiguous terminology is a challenge. "2-way links" would work.

```
    from.ward.asia.wiki.org
    1109 pages with 2-way links
```

Or maybe using your terminology from the pull request.

```
    from.ward.asia.wiki.org
    1109 backlinked pages
```

WardCunningham:
We summarize the size of the neighborhood.
image

Perhaps we should estimate the number of those pages include backlinks.
```
    4130 pages 12% backlinked
```

WardCunningham:
Of course these are just ideas for the future.

Happy New Year.

WardCunningham:
```
if sites[site].sitemap.some(hasLinks)
```
That coded up nicely. (I was worried it was easy to say but hard to do.)

WardCunningham:
Oops. I didn't know what "convert to draft" meant so I pushed the button to find out. Now I need to find out how to un-push that button. This may be tricky since the button disappeared.

WardCunningham:
Let's merge. Paul will push that button.

PR #271, opened 2021-02-27
collaborative links now prefer the title

dobbs:
Prior to this change, doInternalLink() would receive a title or a slug, convert it to a slug, and call showPage() with the slug. refresh.cycle() would be invoked which resolved the slug back into a title. That particular flow would fail for clicks on the graphviz plugin.

This change lets the original title be passed through to createPage for the cases when we already know the intended title, like the graphviz case.


paul90:
Published as `wiki-client@0.23.0`

PR #273, opened 2021-06-29
Highlight errors in the journal

paul90:
Add some highlighting to errors in the journal.

Screenshot 2021-06-29 at 12 26 27

### DIFF
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -31,5 +31,5 @@ module.exports = ($journal, action) ->
       .data("slug", $page.attr('id'))
   if action.error?
     $action
-      .css("border", "2px solid red")
+      .css("border", "2px solid #FFC000")
       .css("margin", "1px")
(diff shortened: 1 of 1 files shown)


WardCunningham:
This is a good middle ground between hiding error messages which we do now and posting a top-of-screen alert banner which we did early on.

Technically this should be a CAUTION because the user has made no mistake. A yellow border is more in line with caution.

It is possible that we can devise repair mechanisms or offer repair suggestions better than "Try reload" which is probably one of the few cases were the message includes a suggestion. Something like the plugin [Help] button could be in order and could link to various journal repair pages. I suggest this here not because it is a prerequisite to merge, but only because it is a good place to write these notes.

Screen Shot 2021-06-29 at 6 37 06 AM

WardCunningham (approved):
I asked for this, didn't I?

paul90:
This is a difficult one...

The time that these errors that appear in the journal really need to be raised is when they happen - just recording them in the journal is effectively just swallowing them. Showing historic errors is almost pointless, in maybe all cases it will result in the page being saved into local storage - so the error is effectively fixed when the page is forked back to the server.

Going to close this one for now.

PR #274, opened 2021-07-30
recognize action.site for all action.type

WardCunningham:
When we drag a node into a Graph from a remote site we note in the Journal were that node can be found.

image


WardCunningham:
Um, I am not sure how I got all of these old commits in this pull request. I changed one line. I also got a notice from github that I should set some upstream parameter which I did. I can throw all of this away and make the change over again if that is the right thing to do.

WardCunningham:
I will try this again in a new pr.

PR #275, opened 2021-07-30
recognize action.site for all action.type

WardCunningham:
redo of #274

### DIFF
--- a/lib/addToJournal.coffee
+++ b/lib/addToJournal.coffee
@@ -21,7 +21,7 @@ module.exports = ($journal, action) ->
     $action.insertBefore(controls)
   else
     $action.appendTo($journal)
-  if action.type == 'fork' and action.site?
+  if action.site?
     $action
       .css("background-image", "url(#{wiki.site(action.site).flag()}")
       .attr("href", "#{wiki.site(action.site).getDirectURL($page.attr('id'))}.html")


paul90 (commented):
This has the desired effect of adding the remote site's flag to the journal action.

But, adding the remote site to `action.site` has the effect of extending the neighborhood. Is this the desired effect? Or should it just extend the context used for links on the page?

This is something I was grappling with back in the drag-attribution branch, and not something that was fully resolved at that time.

dobbs:
Your question reminds me of other conversations we have had about interaction between the journal and neighborhood and resolving links.

I think in this context, the graph plugin will behave like reference plugin. Loading the item expands the neighborhood and that makes the links resolve correctly.

What I think would be somewhat new is that I might drag nodes into one graph item from multiple sites. These would appear as multiple edits with different sites and so one item on the page could expand the neighborhood more widely than a reference—though not more widely than a collection of references.

I continue to think we need another sort of link—something in between a collaborative link and a reference—or perhaps another step in resolving collaborative links. The idea would be to have a link that searches a fixed collection of sites to resolve titles into pages. The neighborhood might only expand after the page has been found.

That suggestion takes us away fro the specific concerns about this PR. Let me return.

I think expanding the neighborhood here is probably fine. As already mentioned it’s pretty close to what we already have with references.

WardCunningham:
Repeating from private conversation ...

We once added a server option that preloaded the neighborhood. I'm not sure it has been used. I can't remember how it worked. It is certainly not on par with Rosters for general utility. Let's think of two different contexts that we expect both program and user to understand. I'll call them search-context and click-context. The search context is the growing neighborhood which can be discarded by refreshing the tab and all links on all pages still work. That's because we expect the click-context to come from the page and to travel with it as a page finds its way around the federation. 

We've recently found the need to add fake-forks to the journal to account for links created by html scripts that look in Rosters for resources. The fake-forks are how we recognize where clickable things came from. Maybe there is a better approach. Maybe we could attach a list of sites to the create Action?

WardCunningham:
The comparison with Reference plugin is informative.

Rendering a Reference grows the neighborhood, the search context. It also keeps the referred-to site in its own Item data and introduces this into the click-context without relying on the neighborhood.

Perhaps the right solution is for Graph to keep a list of sites from which it has been connected so that click-context can be expanded when a node is clicked?

paul90:
Closed after discussion.

Issue #276, opened 2021-10-10
not closing the editor textarea until I choose to do so


rybesh:
The disappear-on-focusout behavior of the fedwiki editor is quite unusual (I can't think of any other webapps that use it). I've always wished I could turn it off as it hinders the flow of editing when switching among windows (as I am almost always doing when editing fedwiki). Instead of just switching back to the browser window and pasting something in, I have to reopen the editor every time.

Furthermore, there are various browser extensions available for using an external editor to edit text in browser textareas, e.g. [GhostText](https://ghosttext.fregante.com). But all of them rely on the textarea continuing to exist when focus switches away from the browser, which is not the case with fedwiki. That's unfortunate, since fedwiki is precisely the kind of application with which one might want to use a fully-featured editor to write text, rather than the impoverished textarea.

Could this design decision be revisited, I wonder?

WardCunningham:
> precisely the kind of application with which one might want to use a fully-featured editor to write text, rather than the impoverished textarea.

I'm confused by this statement because it seems at odds with sharing within a community. We have made a number of decisions to lead folks to make simple pages written in a style that is easy to replicate and make contributions as an equal even when not acquainted with the same full-feature set that you are missing.

On another dimension, we are also strongly interested in ingesting wiki pages as part of workflow automation. Again, raising the bar at the textual level just makes more work reading that text within the automation. This argument is as old as unix where the designers were often criticized for not putting column headings in their output. Their reply was that the headings just get in the way when piping the output of one program into the input of another.

PR #277, opened 2021-10-22
Don't close the editor on losing focus

3-w-c:
Ref #276

By @Bortseb and others

Seems To Work™

Test plan:

1. Run grunt to compile client JS (let me know if the resulting JS should be added in-repo as well)
2. Run fedwiki
3. Edit a paragraph
4. Click away
5. The editor should stay open
6. Try and open another editor, it should refuse to until the initial editor is closed

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -34,12 +34,12 @@ textEditor = ($item, item, option={}) ->
 
     if e.which == 27 #esc for save
       e.preventDefault()
-      $textarea.focusout()
+      focusOutHandler()
       return false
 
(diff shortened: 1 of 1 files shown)


WardCunningham:
I recommend folks who desire this functionality use this actively for a month or two and report no adverse consequences. At this point a case can be made to other frequent users that this is better focus out behavior. We do often suggest that users save by clicking out rather than figuring what keystroke they might use on their keyboards. We should stop proposing focus out and suggest ESC instead. 

Passing this hurdle, we need to inspect the context in which this code runs to be sure that there is no case where action updates cascade and that edited text is saved reliably in all cases.

PR #278, opened 2021-12-27
Assign cloned item to  when aliasing to prevent undesired ui changes

andrewshell:
Steps to replicate:

1. Have two pages open next to each other
2. Create a new paragraph in the first column
3. shift-click-drag the paragraph to the second column (to copy)
4. double-click the original paragraph and change the text

What I expect to happen:

1. Only the text in the first column changes

What actually happens:

1. The text changes in both columns (only in UI, if I refresh the page it corrects itself)

Resolution:

1. In `refresh.coffee` function `aliasItem` we are already cloning `oldItem` as `item` (via `$.extend`)
2. Assign this cloned `item` to `$item` via the data method (`$item.data('item', item')`)

### DIFF
--- a/lib/refresh.coffee
+++ b/lib/refresh.coffee
@@ -35,6 +35,7 @@ getItem = ($item) ->
 
 aliasItem = ($page, $item, oldItem) ->
   item = $.extend {}, oldItem
+  $item.data('item', item);
   pageObject = lineup.atKey($page.data('key'))
   if pageObject.getItem(item.id)?
     item.alias ||= item.id


WardCunningham:
Thank you for working out this fix in some well-worn code and including clear replication instructions.

Note: aliasItem serves another purpose but seems as reasonable as any other place to attend to this edge case.

andrewshell:
@WardCunningham You mention in [End Drag Refactoring](http://ward.asia.wiki.org/view/end-drag-refactoring):

> Let's embrace footer item search as our best method of exploring provenance by getting rid of shift-drag and adopting this simple rule: drag within a page is move, between pages is copy.

Since I just reviewed that code and it's fresh in my mind, would you like me to implement this change?

WardCunningham:
Yes, this code is overdue for reconsideration. There are some quirks in the work that Joshua did that didn't show up immediately. The one that comes to mind is there are restrictions on dragging from identically named pages. This occurs in several circumstances. I know I have copied content to a temporary location just to get around this restriction.

- copying new content from a page previously forked -- desirable
- copying content from a previous revision of a page -- desirable
- copying content from a single page viewed twice in the lineup -- undesirable

I suppose there are many other cases that should be considered including cases that motivated the alias mechanism.

I am having some trouble figuring out what I thought this has to do with provenance. Have you deciphered that?

andrewshell:
No, I wasn't sure about provenance. I figured I'd try to hop on the call tomorrow and ask.

paul90:
Thank you Andrew

PR #279, opened 2021-12-28
RFC: Adding a presave hook in editor focusoutHandler

andrewshell:
I'm curious if there have been any conversations about providing hooks in the code to allow plugins to extend core functionality.  Here is an example of adding a "presave" hook to the editor.

I've created a plugin that would use this hook as a proof of concept.

https://github.com/andrewshell/wiki-plugin-metadata/blob/master/client/metadata.js#L26-L35

I'm sure there will need to be discussions before something like this gets merged. So I'm documenting the idea with working code.

### DIFF
--- a/lib/editor.coffee
+++ b/lib/editor.coffee
@@ -95,7 +95,9 @@ textEditor = ($item, item, option={}) ->
     $item.removeClass 'textEditing'
     $textarea.unbind()
     $page = $item.parents('.page:first')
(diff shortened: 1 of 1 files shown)


dobbs:
A while back I created a feature to allow plugins to override the editor. It's not the same thing as the presave hook you've proposed, but I think reaching for similar kinds of extensions.
https://github.com/fedwiki/wiki-client/pull/247

PR #280, opened 2021-12-28
Enable wiki URLs to declare a lineup using location.hash in addition to location.pathname

dobbs:
# Goals
Enable wiki-client to support a statically hosted wiki.
Preserve support for dynamically hosted, node-based wikis.

# Rationale
We introduce an internal structure to represent the Reader's intent for the lineup. We believe this opens several opportunities.

We enables a URL hash parameter to provide the intended lineup while preserving support for existing use of URL pathname. Using URL hash instead of URL pathname opens the door to statically hosted wikis.

We anticipate this intermediate structure enabling other changes to clarify responsibilities between state and lineup.

The {site, slug} invites {site, slug, revision} and could open the door for better support of non-European languages with something like {site, slug?, revision, title}

We notice that using the URL hash invites collisions. Google are experimenting in a recent chrome plugin to enable the use of  `#:~:` to delimit links to text fragments. This PR makes no attempt to detect or resolve such collisions.

One other decision hidden in this code is the choice to use the specific parameter "pathname" to hold the intended lineup. This opens experimenting with other URL structures like the slug@site/slug@site we have in Seran Outpost and other experiments.


lib/state.coffee
```diff
@@ -73,3 +114,11 @@ state.first = ->
...
     link.createPage(urlPage, firstUrlLocs[idx]) unless urlPage is ''
 
+state.syncDomWithLocation = ->
+  main = ""
+  for {site, slug} in state.fromLocation(location)
+    dataSite = ""
+    if site != "view"
+      dataSite = """data-site="#{site}" """
+    main += """<div id="#{slug}" #{dataSite}class="page"></div>\n"""
+  $('section.main').html(main)
```
dobbs, line 124: This function is where we introduce more confusion to the bootstrapping of wiki-client.

We replace the contents of `<section class="main">` with our own page divs computed from `window.location`.

For comparison, here is where @paul90 made a similar intervention in the DAT/hypercore variant.
https://github.com/paul90/wiki-client-dat-variant/commit/15cb65667861abd9c9af80d7970b43f1c001246b#diff-5b6ac0ea3ef5483a6e071c1ff85d99a3c4ef31d97b567deff46686fb93a25921R17-R33

For the time being, we are deferring the work to understand and change the existing wiki-client bootstrap.

test/state.coffee
```diff
@@ -0,0 +1,91 @@
...
+    ['/view/welcome-visitors/fed.wiki.org/featured-sites',
+      ['view', 'fed.wiki.org'], ['welcome-visitors', 'featured-sites']],
+    ['/welcome-visitors.html', ['view'], ['welcome-visitors']],
+    ['/how-to-wiki.html', ['view'], ['how-to-wiki']]
+  ]
+  before ->
+    global.$ = (el) -> {attr: (key) -> el[key]}
+    global.history =
+      pushState: (state, title, url) -> actual = url
+    lineup.bestTitle = () -> title
```
dobbs, line 24: We're stubbing jQuery's `$`, `history`, and the imported `lineup`. In specific test cases below we also stub `location`, and `document.title`. These allow the tests to pass where we do not have these browser APIs, but they will also be too tightly coupled to the current implementations.

dobbs (commented):
A few notes about some trade-off decisions we know about.

dobbs:
We have a companion to this PR in a separate repo: [readonly.js](https://github.com/dobbs/deploy.wiki.do/blob/738f7e41b3cbe2b1bebac0e6afc9cacb40153503/data/assets/bin/readonly.js)

This javascript program copies client side code and pages into a folder suitable for publishing with github pages.

There is some chance we'll include the publishing script in this PR before we're done.

WardCunningham:
Let's consider for a moment what one must do to be sure that a hash url saved will continue to work in the presence of url format evolution.

First we note that some server will provide the wiki-client that must recognize any particular url and the name of that server is already recorded in the saved url. This gives us considerable latitude.

However, if a server operator chooses to vary the scheme for sites made public, they have some obligation to support all variations going forward. This seems a modest burden but one the authors of an open-source wiki-client must bear.

It is possible that the url scheme could include some page specific information. We could, for example, provide a ghost page with some information currently lost to the url. For example, a search result could include the search term in the url so that the search is easily repeated. Maybe the hash is base 64 json?

Continuing in this direction, Wiki-21 has suggested developing the "panel" abstraction. The hash could then consist of a rollup of plugin, panel and lineup state to be reconstituted at will.

These opportunities are beyond this pull request unless some thought experiment along these lines suggests how we might enable this future.

PR #282, opened 2022-01-06
Client updates

paul90:
Update the client add some updates made on the server, so the sitemap and index in the current session reflect the changes made on the server.

* References are included in the links in the sitemap - fedwiki/wiki-server#169
* The URLs from external links are excluded from the text being indexed - fedwiki/wiki-server#170


WardCunningham:
Very nice. Thank you.

PR #286, opened 2022-03-27
likely change for drop-import of single page-json file

WardCunningham:
We have had a request for the ability to share a single wiki page with other federated wiki users. A suggested application what selectively sharing private pages via other file sharing such as email.

We note that the JSON link at the bottom of every page will respond favorably to the `Save Link As...` menu by writing a page-json page for any page visible to the current viewer, logged on or not.

This modification notices when a single page-json file has been dropped as if it were export json. We wrap the dropped page with another layer of object so that it falls into line with normal import. We also adjust the importer page synopsis to make it clear how the drop has been interpreted.


paul90:
Included in + wiki-client@0.23.3

PR #287, opened 2022-04-19
correctly count number of pages

paul90:
Pages count was getting increased incorrectly each time the sitemap got updated - adding the number of pages in the origin site each time.

This change corrects that, by summing the number of pages from the sitemaps for the sites in the neighbourhood.


WardCunningham (approved):
Pair programed together.

paul90:
Published as + wiki-client@0.23.4

PR #288, opened 2022-06-09
update url on fork from history

WardCunningham:
First step in lineup reconciliation when forking from history.
Here we update the url to remove the _rev suffix from the page that is now the current version.


paul90 (commented):
See https://github.com/fedwiki/wiki-client/pull/288#issuecomment-1159047020

paul90:
Published as + wiki-client@0.23.5

PR #289, opened 2022-06-13
ignore any sites with sitemap request still inflight when counting pages

paul90:
The total number of pages gets recalculated triggered by the `new-neighbor-done` event. However, some neighbors might not be done yet so those will fail with a type error when trying to count the number of pages with `site.sitemap.length`.

In this change any site that are still in flight, `sitemapRequestInflight` is still `true`, are ignored. They will eventually get included as they complete and fire their own `new-neighbor-done` event.

### DIFF
--- a/lib/neighbors.coffee
+++ b/lib/neighbors.coffee
@@ -46,7 +46,16 @@ bind = ->
       catch error
         console.info '+++ sitemap not valid for ', site
         sites[site].sitemap = []
-      totalPages = Object.values(neighborhood.sites).reduce ((sum, site) -> site.sitemap.length), 0
+      totalPages = Object.values(neighborhood.sites).reduce ((sum, site) -> 
+        try
+          if site.sitemapRequestInflight
+            return 0
+          else
+            return site.sitemap.length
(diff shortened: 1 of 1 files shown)


WardCunningham (commented):
I wonder, have you seen a failure that leads to the try-catch around what seems like rather safe code?
When the unexpected happens isn't it often easier to debug when the browser catches the error at the source?

paul90:
I spotted this problem both when loading a page that added more than one site to the neighbourhood, and refreshing the browser when the neighbourhood had 2 or more sites.

I think the catch is going to be needed if there is something very wrong with a site. The `sitemapRequestInflight` event is fired, and `sitemapRequestInflight` unset, when a sitemap has loaded. So, by ignoring sites where the request hasn't completed there should be a something in the sitemap, and the catch will only happen when it doesn't have a length.

WardCunningham:
Ah. A condition that I hadn't considered. Thanks for explaining this. I'm happy to see it merged.

paul90:
published as + wiki-client@0.23.5

PR #290, opened 2022-06-18
remember to add number of pages to the running total

paul90:
At some point I there was the change from `totalPages += pageCount` on each sitemap load, which if I remember correctly was inflating the number. To doing the count across all the sitemaps from fresh each time, using `reduce()`. For whatever reason that reduce never added the number of pages in each wiki into the running total. 😊


WardCunningham:
This is adding up the page counts consistently.
The number agrees with an older version of wiki when it is behaving well.

image

My wife gets the same result when dictating the individual site page counts into her wrist watch.
I'd say this is good to go.

paul90:
Publlished as + wiki-client@0.23.6

PR #291, opened 2022-08-02
ghost rename, work in progress

WardCunningham:
Work To Do
========
- Be sure the text is clean. Especially avoid accidental copy-paste of html
- Address the issue of deleting renamed versions of pages. (Recalling 'create' resets the title.)

Features So Far
===========

image


paul90:
A number of changes got clobbered by that 'merged with master' commit.

WardCunningham:
Robert Best and I have reviewed this pull request and would like to do the work toward completion together as follows:

- Be sure the text is clean. Especially avoid accidental copy-paste of html -- probably just some regex, revert if unacceptable
- Remove _rev0 from id and url (and more?) when renaming a journal create item -- do the suggested removes, more test cases
- Page names in Assets folders. -- we should not mess with assets, good example case is Eric duplicating a page with code to explore a new variation while using esm.html

I'm unsure about how that last commit did nothing useful and was so damaging. How can we remove it from this pull request? Please help.

paul90:
Well this PR was started back in 2022, so there have been changes.

Rather than trying to merge master into these changes, it would be better to rebase. So,

1. do a hard reset to get rid of that merge.
2. with ward/title checked out `git rebase master`

There appears to be two areas of conflict. The only one that needs any attention is in `lib/addToJournal.coffee`. For `package-lock.json` pick the one in master.

Once all is resolved, you will need to force push the update back to github.

WardCunningham:
Paul and I fixed up this pull request which was against an old version and merged incorrectly. It's in good shape now and ready for finish work called out above.

WardCunningham:
Replaced with #342

PR #292, opened 2022-08-16
Client Update (Jul/Aug 2022)

paul90:
Client updates associated with [image plugin](https://github.com/fedwiki/wiki-plugin-image/pull/6) update, and some other fixes.

The first part of this PR, contains the early work on the image plugin custom editor, which was mostly removed to the image plugin in c10a83e9ffb16f3c4b0601cd4653d842f1871c2e

The remainder is a revival of "Attribution on Drag" from some years ago. Finally added to the client so that when an image is dragged between sites the image plugin has a chance of finding the image if it has not also been forked.


Gruntfile.js
```diff
@@ -58,7 +58,7 @@ module.exports = function (grunt) {
         src: ['./client.coffee'],
         dest: 'client/client.max.js',
         options: {
-          transform: [['coffeeify', {transpile: {presets: ['@babel/preset-env']}}]],
+          transform: [['coffeeify', {transpile: {presets: ['@babel/preset-env'], plugins: ['@babel/plugin-transform-runtime']}}]],
```
WardCunningham, line 61: We were already using Babel? Now using a runtime? What should programmers know about this?

lib/image.coffee
```diff
@@ -45,7 +45,10 @@ bind = ($item, item) ->
       item.source
     else
       item.url
-    dialog.open item.text, """<img style="width:100%" src="#{url}">"""
+    dialogTitle = (item.text||'').replace /\[\[([^\]]+)\]\]/gi, ''
```
WardCunningham: What is the goal here? Does this just keep some markup out of the dialog title?

paul90: This removes any links included in the image caption so they don't appear in the dialog's title.

PR #293, opened 2022-11-24
updated time on action hover

WardCunningham:
Old hover:
Screen Shot 2022-11-24 at 12 14 25 PM

New hover:
Screen Shot 2022-11-24 at 12 05 18 PM

Recompute time on each new hover.

We won't see actions claiming to have occurred milliseconds ago for long.


WardCunningham:
Suggest we move the action title logic to util.coffee.
I'll add that to this pull request.

paul90:

PR #295, opened 2022-12-06
give search results ghost a journal

paul90:
adding a journal with a create actions to the search results allow a forked result page to be later removed.

### DIFF
--- a/lib/search.coffee
+++ b/lib/search.coffee
@@ -156,9 +156,6 @@ createSearch = ({neighborhood})->
         "story": deepCopy(resultPage.story)
       },
       "date": Date.now()
-    },{
-      "type": "fork",
-      "date": Date.now()
     }]
     pageObject = newPage resultPage
     link.showResult pageObject
--- a/lib/search.coffee
+++ b/lib/search.coffee
@@ -14,6 +14,9 @@ page = require './page'
 escapeRegExp = (string) ->
   string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 
+deepCopy = (object) ->
+  JSON.parse JSON.stringify object
+
 # From reference.coffee
 emit = ($item, item) ->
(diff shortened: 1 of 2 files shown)


lib/search.coffee
```diff
@@ -126,22 +129,39 @@ createSearch = ({neighborhood})->
...
+      "type": "create",
+      "item": {
+        "title": resultPage.title,
+        "story": deepCopy(resultPage.story)
+      },
+      "date": Date.now()
+    },{
+      "type": "fork",
+      "date": Date.now()
+    }]
```
WardCunningham, line 162: I'm trying to remember the logic for adding this extra fork. Our delete logic is based on viewing the first revision recalled from history. A page with only the create in the journal is not subject to the delete logic.
https://github.com/fedwiki/wiki-client/blob/f3f7043ca09456e01b6331abdea9b9e916ed265d/lib/legacy.coffee#L242

I can remember having some trouble that was resolved by adding a fork from the ui such that some subsequent operation was made visible. I can't put my finger on it. But I don't think anticipating this situation and adding the extra action is necessary. Here are the normal cases that will work fine without it.

- ghost page with create only appears
- ghost page is abandoned with no impact on the server
- ghost page is forked to live on in the server, now has two actions in the journal
- page living on in the server is to be discarded, open create action and fork, done.

paul90: > A page with only the create in the journal is not subject to the delete logic.

Only this is not what happens to a ghost page with only a create action.

WardCunningham (approved):

WardCunningham:
[f07a625](https://github.com/fedwiki/wiki-client/pull/295/commits/f07a6257fda9640176f2cea66080dd8fe0b70f1b) Solves a problem where the `_rev0` suffix appended to a url made a fork look like a delete.

paul90:

PR #296, opened 2022-12-20
sitemap refresh

paul90:
Add a periodic refresh of sitemaps, if there have been any remote changes.

Add some extra information to the hover information on the flags in the neighbourhood bar.
Screenshot 2022-12-20 at 18 54 02

The refresh interval depends on how long ago the sitemap was updated. The initial thought was that they should be checked as far in the future as they were last updated in the past, but limited to be no more than a day. This is probably too infrequent, so will most likely be reduced.


WardCunningham (approved):

paul90:

PR #297, opened 2023-01-13
Security: Update json5 module

heymatthew:
These changes were made using `npm audit fix`.

When running npm audit against master, you get this warning:

> Severity: high
> Prototype Pollution in JSON5 via Parse Method

You can read about it more here: https://github.com/advisories/GHSA-9c47-m6qq-7p4h

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -580,18 +580,6 @@
         "node": ">=4"
       }
     },
-    "node_modules/@babel/core/node_modules/json5": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.1.tgz",
-      "integrity": "sha512-1hqLFMSrGHRHxav9q9gNjJ5EXznIxGVO09xQRrwplcS8qs28pZ8s8hupZAmqDwZUmVZ2Qb2jnyPOWcDH8m8dlA==",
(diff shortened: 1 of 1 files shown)


paul90:
Thank you, but this is not the way to fix a problem that is within the dependencies of a package we rely on as part of our development build chain.

Look at were it is used:
```
❯ npm ls json5
wiki-client@0.24.2 
└─┬ @babel/core@7.19.3
  └── json5@2.2.1
```

If we remove `package-lock.json` and the `node_modules` directory, and re-install the problem is fixed.
```
❯ npm ls json5
wiki-client@0.24.2 /Users/Paul/Projects/FedWiki/wiki-client
└─┬ @babel/core@7.20.12
  └── json5@2.2.3
```

There was a change in [@babel/core@7.20.12](https://github.com/babel/babel/releases/tag/v7.20.12) to address the security issue with `json5`, so we can update to use the updated @babel/core.

As a rule this is a better approach that forcing a version in `package-lock.json` for those packages that are depended on by those we depend on.

paul90:
The version of @babel/core and its dependencies are updated in 85ccd84fd044832d2ae6f7f432e2ce8a6a2e8c89

heymatthew:
Thanks @paul90

PR #300, opened 2023-01-18
Update to minisearch 6

paul90:
This update started as a fix for an issue on the server with updating the search index when a page is forked over an existing page. While investigating that issue an error was spotted with the regex that removes HTML tags.

This PR includes:
* a correction to the regex used to remove HTML tags,
* a simplification of the code made possible by minisearch@6, which removes the need for a pages original text when removing it from, or updating it, in the index.


WardCunningham (commented):
Can you say a few words about the edits in lib/pageHandler.coffee? It looks like the diff got a little carried away since many of the red and green lines look identical.

What is it about search that required modification here?
I think there are a bundle of odd cases handle by that code so I wonder it deserves careful testing? If the change is truly surgical, then maybe not.

WardCunningham (approved):
Awesome

PR #305, opened 2023-07-26
update jquery version in runtest

paul90:
Found the reference to the old version of jQuery that was causing Dependabot to raise an alert. As the version of jQuery being referenced is not included here, so the real problem would be that `runtest.html` would not have worked.

### DIFF
--- a/client/runtests.html
+++ b/client/runtests.html
@@ -4,9 +4,9 @@
     <meta charset="utf-8">
     <title>SFW Mocha Tests</title>
     <link rel="stylesheet" href="/test/mocha.css" />
-    <script src='/js/jquery-2.2.4.js' type='text/javascript'></script>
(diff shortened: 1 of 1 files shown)


WardCunningham:
I have the faintest memory of asking for an online test runner when mocha was introduced. I can't remember when I last looked at it. This week I find myself all excited about online client tests. Perhaps I am overly enthusiastic?

I gave it a try on the nearest open window, own.fed.wiki, hosted on the reasonably current asia.wiki.org. It didn't seem to work, as you said.

image

paul90:
updated, all back to normal.

Screenshot 2023-07-29 at 12-06-29 SFW Mocha Tests

WardCunningham:
Wow. That looks really nice. I'm less proud of my most recent testing creation.

PR #306, opened 2023-08-21
Adding support for Login to View

paul90:
Some changes related to Login to View.

Refactoring the sitemap handling to use siteAdapter, so we use its support for login to view, and adding support for returning lastModified date when returning sitemaps.

Adding 'login required' support for sitemaps - these will often be the second resource fetched from a remote wiki, after the flag.

### DIFF
--- a/lib/siteAdapter.coffee
+++ b/lib/siteAdapter.coffee
@@ -348,7 +348,7 @@ siteAdapter.site = (site) ->
           url: url
           xhrFields: { withCredentials: useCredentials }
           success: (data, code, xhr) ->
-            if data.title is 'Login Required' and !url.includes('login-required') and credentialsNeeded[site] isnt true
+            if ((route is 'system/sitemap.json' and Array.isArray(data) and data[0] is 'Login Required') or data.title is 'Login Required') and !url.includes('login-required') and credentialsNeeded[site] isnt true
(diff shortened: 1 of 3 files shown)


lib/siteAdapter.coffee
```diff
@@ -158,7 +158,11 @@ siteAdapter.origin = {
       type: 'GET'
       dataType: 'json'
       url: "/#{route}"
-      success: (page) -> done null, page
+      success: (page, code, xhr) ->
+        if route is 'system/sitemap.json'
+          done null, { data: page, lastModified: Date.parse(xhr.getResponseHeader('Last-Modified') )}
```
WardCunningham, line 163: We are guessing if a lazy origin implementation didn't attach the header then elsewhere you detect NaN and call that zero.

WardCunningham (approved):

lib/siteAdapter.coffee
paul90: This is handled in [neighborhood.coffee line 51](https://github.com/fedwiki/wiki-client/blob/42e9af8ecca71989f2539ba10521a6b4498f939b/lib/neighborhood.coffee#L51), part of this PR.

paul90:
Published as + wiki-client@0.27.0, included in + wiki@0.31.0

PR #307, opened 2023-09-24
Address the forever spinning flag when wiki is unreachable.

paul90:
Addressing the ever spinning flag, caused by the data destructure failing, so the error handling is never reached.

Error message from console:
 Uncaught TypeError: e is null
    u client.max.js:1978
    r client.max.js:5035
    get client.max.js:5095
    u client.max.js:1976
    n client.max.js:2028
    setTimeout handler*r client.max.js:2035
    registerNeighbor client.max.js:2053
    load_sites welcome-visitors line 2 > injectedScript:42
    bind welcome-visitors line 2 > injectedScript:186
    jQuery 2


WardCunningham (approved):
Oh, I hadn't noticed the ever spinning. Now I never will.

PR #308, opened 2023-10-10
When implicitly forking a page, include the page.

nrn:


paul90:
Have both explicit and implicit fork use `forkPage` to hold the page being forked. Will hopefully aid comprehension when we look at this, and the associated server code in some years time.

WardCunningham:
This change in the client to server message format is an improvement that depends on us deploying client and server together. Merge at will to be installed as a new wiki, not client or server individually.

paul90:
As discussed on call, adding the date to the implicit fork event - this value stays on the client, but ensures that it is preserved if the fork is repeated without reloading the page.

paul90:
Published as + wiki-client@0.28.0

PR #311, opened 2024-02-08
Replacing jQuery UI Dialog

paul90:
Replacing jQuery UI dialog with a child 'popup' window. By changing `dialog.coffee` this replaces all jQuery UI based dialogs at once.

On some browsers and devices the child window might open in a tab, or if the user has the opening of child windows blocked be requested to unblock for the wiki site. But, it removes the dialog that obscures a portion of the screen, and may present a challenge in a small windows. 

Replacing Screenshot 2024-02-08 at 10 38 12

with Screenshot 2024-02-08 at 10 41 03


client/dialog/index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title></title>
+    <link id="favicon" href="/favicon.png" rel="icon" type="image/png" />
+    <link href="/dialog/style.css" rel="stylesheet" />
+    <script>
+      window.addEventListener(
+        "message",
```
WardCunningham, line 9: Is this the only message that the dialog will handle now and into the future?
If the message will eventually include a function code, perhaps that should be established now?

paul90: This PR is just replacing the existing `wiki.dialog` behaviour of displaying some static content in a dialog window.

PR #312, opened 2024-02-15
Update Text Extract For Indexing

paul90:
Update the text extract logic to include better removal of markup tags. This mirrors the change made on the server in fedwiki/wiki-server#183


WardCunningham:
Sweet. 
Thank you.

Issue #284, opened 2022-02-04
Wiki page missing, yet its content shows up in wiki search...


cliveb:
Any idea why a wiki page goes missing?

http://clive.tries.fed.wiki/view/recent-changes/view/teradata-v-sap

The page content shows up in search. 

I am using a Chromebox, as I normally do writing on wiki, with the latest ChromeOS.
![page missing found in search](https://user-images.githubusercontent.com/733870/152479884-611322e0-d949-4f0e-8f1c-9c3a561d9bdf.png)

I write this page from about 1 am to 3 am PST Feb 3, 2022.

Thanks for any pointers.

cliveb:
Update. I found the page from the wiki search, by clicking the color wiki icon square in the browser (not the blue link in the browser). The page loaded, in the open window, I forked the page, but the only the page paragraph in the observable in the wiki browser forked. 

Is this my user error. 

I noticed another inconsistency (not 100 percent sure) I found I could fork the page without signing to my wiki account. Again the fork was only the paragraph seen in the wiki search browser preview. 

Any chance the wiki server DB has reached its allocated size limit, and is now writing beyond memory? A similar pattern infamously happened to FourSquare whose database had disk space but the space was not allocated in DB, which caused the index to be searchable but the document pages were not found (I think 4sq was on mongodb)

.  
![forks only wiki search paragraph](https://user-images.githubusercontent.com/733870/152644162-2f681b8f-ef82-44f6-b9ff-aef3c83b2d9f.png)

paul90:
Interesting...

Looking at the list of pages, I see the slug is `teradata-v-sap-`. So, I suspect the title has a trailing space. On the "Deep Thoughts" page the link as a trailing space in the link.

You ask, "User error?" - well that may be, but it uncovers some inconsistency in how spaces in links, titles, and slugs are trimmed. Which will need investigating.

In the sitemap we see,
```
{
  "slug":"teradata-v-sap-",
  "title":"Teradata v SAP",
  "date":1643884690759,
  "synopsis":"A good example of a So...186/603 docket]"
}
```
which suggests that the trailing space in the title has been trimmed off at some point. Which creates a mismatch between the slug generated from the title, and the slug used to save the page - which almost certainly explains the other strangeness.

cliveb:
Yes. Even wiki page edits and page forks are all malformed. The trailing space appears to cause all the strangeness. 
[...]

PR #314, opened 2024-02-19
Handle internal links that start/end with spaces

paul90:
Addressing an old usability trap (#284): internal links that start/end with spaces look much the same as one that doesn't, but creates a different name for the wiki page file.

[[space]] -> 'space',
[[ space]] -> '-space'

Simply trimming the space off of the internal link is not a solution, as it would leave those pages that already exist, with such names, as unreachable.

In this change we add some styling to highlight internal links that start/end with a space. Screenshot 2024-02-19 at 10 49 06 

On the "_We could not find this page._" page, the create page line is replaced with a message. _A better message is needed._

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -12,7 +12,7 @@ a {
   text-decoration: none; }
 
 /* highlight links that have slug starting or ending with a `-`, a space in the page title */
-a[href^='/-'], a[href$='-.html'] {
+a.spaced {
   text-decoration: wavy underline crimson;
 }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -11,6 +11,11 @@ pre.error {
 a {
   text-decoration: none; }
 
+/* highlight links that have slug starting or ending with a `-`, a space in the page title */
+a[href^='/-'], a[href$='-.html'] {
+  text-decoration: wavy underline crimson;
+}
+
 a img {
   border: 0; }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -11,8 +11,8 @@ pre.error {
 a {
   text-decoration: none; }
 
-/* highlight links that have slug starting or ending with a `-`, a space in the page title */
-a.spaced {
+/* highlight links that have title starting or ending with a space. */
+body:has(#security>a>i.fa-lock-open) a.spaced {
   text-decoration: wavy underline crimson;
 }
 
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -12,7 +12,7 @@ a {
   text-decoration: none; }
 
 /* highlight links that have title starting or ending with a space. */
-body:has(#security>a>i.fa-lock-open) a.spaced {
+body:has(footer span[style=""]) a.spaced {
   text-decoration: wavy underline crimson;
 }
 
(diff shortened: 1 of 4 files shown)


almereyda:
Thanks for the improvement of this particular case. Indeed that happens.

To comment on the styling presented in the screenshot: While I can see the space after the second link none of the three links appears to begin with a space.

The red curly line might also be mistaken for the spell checker. Would it seem useful to display it in blue, too?

WardCunningham:
The original wiki would test each link as it was rendered and show it with a question mark. In wikipedia this became the red link. I chose to forgo this affordance in federated wiki because the check seemed overly complex and unreliable. However, authors linking to pages that can't be resolved in the page context seems a much more common problem than the unintended leading space. I fear we are establishing a precedent for link evaluation which is more syntactic than semantic, the latter being the more common mistake.

I also note that a trim() in the Create New Page handler (In the HTML plugin) would be a tiny modification with no upward compatibility issues. On our call last Sunday, when Carolina asked if adding a trim() in the right place would be a good first commit I was about to say yes, thinking of the form handler. 

I realize I'm mentioning two "could have" situations rather than examining the pull request carefully. Please be patient with me. I'd like to think the tradeoffs present here a little longer. Thank you.

paul90:
It goes without saying that this is a thorny problem. Adding `trim()` for the create new page form handler was unlikely to cause problems. Being performed from a form field, there was no page link that would be left pointing to the wrong place.

Removing the offer to create a blank page on the ghost missing page is probably the safest approach - preventing the creation of pages with titles that start, or end, with a space. Trimming the spaces, and offering to create a new page is not an option as it would still point to a missing page even if a new page was created. In the same way as trimming links could leave some existing pages unreachable.

On reflection I'm not sure about the highlighting of these links. Maybe it should only be visible to the wiki owner.

paul90:
That list isn't quite correct, it will also include pages with titles that start with one, or more, characters that the slug process removes followed by a space. Or such a sequence at the end of the title. 

To get the full picture this also needs to look at the titles.

So, while there are titles on found.ward.bay.wiki.org that start with a space, such as " Discovery - BBC World Service"  ("-discovery---bbc-world-service"), there is "λ-Calculus Keeps On Giving" ("-calculus-keeps-on-giving").

Using `sitemap.filter((e)=>e.slug != wiki.asSlug(e.title.trim())).map((e)=>e.title)` gives a list of just the pages titles which start or end with a space.

client/style/style.css
```diff
@@ -11,8 +11,8 @@ pre.error {
 a {
   text-decoration: none; }
 
-/* highlight links that have slug starting or ending with a `-`, a space in the page title */
-a.spaced {
+/* highlight links that have title starting or ending with a space. */
+body:has(#security>a>i.fa-lock-open) a.spaced {
```
paul90, line 15: There are some security schemes that don't have the lock, might be better to use the wiki tick - `body:has(footer span[style=""])`

paul90:
From looking at using the page title, with spaces, one the ghost page.

It looks as if a [change for internal links from graphviz](https://github.com/fedwiki/wiki-client/blame/8fbad0152f613ae72f4fed5b6079be7c26dda28f/lib/refresh.coffee#L468) trims the title, so in the future plugin the title no longer has the spaces that were there when the slug was created.

PR #315, opened 2024-02-26
fix: handle case when site has a port number in URL

paul90:
As well as removing any subdomains, also remove any port before checking if the remote is a wiki on localhost.


WardCunningham:
Thanks for this.

PR #316, opened 2024-03-01
Fix broken plugin dialog

paul90:


WardCunningham:
This seems to be missing the error message that we once put in the title/heading of the jQuery dialog.
Maybe that text would be best treated as an H3 or similar?

WardCunningham:
Thanks. Merge incouraged.

PR #319, opened 2024-04-22
Add missing item id

paul90:
The reference items on the search results ghost page were missing an itemId. This causes a problem elsewhere.

Thanks @dobbs for raising this during Sunday's call.


WardCunningham (approved):
Thank you for this.

PR #320, opened 2024-06-18
Use dynamic import to load plugin client code

nrn:
I know it seems insane, but this seems to just work.

By using the built in import expression we can dynamically load scrips like we have been, but those scrips in turn can now use import statements and easily load their own dependencies without a build step. For a plugin that would often mean pulling from its own plugin folder like
```js
import { test } from '/plugins/mech/test.mjs'
```

The grunt change is required to prevent babel from replacing native dynamic imports with it's own stuff. This is the only call to loadScript so it was easy to cut the unused options argument, and convert the response to normal promise methods.

This is going to require a lot of testing.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,7 @@ module.exports = function (grunt) {
         src: ['./client.coffee'],
         dest: 'client/client.max.js',
         options: {
-          transform: [['coffeeify', {transpile: {presets: ['@babel/preset-env'], plugins: ['@babel/plugin-transform-runtime']}}]],
+          transform: [['coffeeify', {transpile: {presets: [[ '@babel/preset-env', { "exclude": ["proposal-dynamic-import"]}]], plugins: ['@babel/plugin-transform-runtime']}}]],
           browserifyOptions: {
             extensions: ".coffee"
           }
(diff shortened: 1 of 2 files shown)


WardCunningham:
I wonder how far back in browser versions we would have to go before this won't work? On a truly obsolete browser one would still have access to the built-in plugins. This would be a feature simplifying the life of new plugin authors and certainly a step in the right direction.

paul90:
A number of plugins already import their own dependencies, without a build script. Albeit using a dynamic import, like `const Plot = await import('https://cdn.jsdelivr.net/npm/@observablehq/plot@0.6/+esm')`.

But switching to static import would simplify importing things that are always needed, but dynamic is probably probably preferred for things that are not always needed.

Should work in quite old browsers, https://caniuse.com/?search=import(),

nrn:
Yeah, that's a good point, we could use dynamic imports in plugin code since they can be used from either scripts or modules in the browser, or in node from CJS or .jsm, and can cross load any of them from any situation. This just seems like a really convenient boundary to put dynamic imports at so we can use import statements to simplify things in plugin land.

paul90:
Those plugins which themselves are using dynamic plugins fail with this change.

nrn:
Aargh, caught out by strict mode compatibility! It's an easy fix in the `code` plugin, but I'm not sure how many might be a problem.

paul90:
Finally got around to a rc for the change to the code plugin - it is installed on the test server. The only update not installed there is the change to the graphviz plugin.

paul90:
Merging as discussed in Sunday's call.

PR #321, opened 2024-07-11
Convert plugin loading to async/await forward JS

nrn:
The main reason for this conversion is to contain errors on a per plugin basis so one broken/missing plugin doesn't break the rest of the page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "wiki-client",
-  "version": "0.29.2",
+  "version": "0.30.0-rc.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "wiki-client",
-      "version": "0.29.2",
+      "version": "0.30.0-rc.1",
       "license": "MIT",
       "dependencies": {
         "async": "^3.2.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wiki-client",
-  "version": "0.29.2",
+  "version": "0.30.0-rc.1",
   "description": "Federated Wiki - Client-side Javascript",
   "keywords": [
     "wiki",


lib/plugin.js
```diff
@@ -130,72 +128,80 @@ const bind = function(name, pluginBind) {
...
   if (window.pluginSuccessor[name]) {
     wiki.log('plugin successor', name, window.pluginSuccessor[name]);
     name = window.pluginSuccessor[name]; 
   }
-  if (loadingScripts[name]) { return loadingScripts[name].then(callback); }
-  loadingScripts[name] = new Promise(function(resolve, _reject) {
-    if (window.plugins[name]) { return resolve(window.plugins[name]); }
-    return getScript(`/plugins/${name}/${name}.js`, function() {
-      let p = window.plugins[name];
+  if (window.plugins[name]) {
```
dobbs, line 169: This might be something that was already there in the previous code...

There are a lot of guard clauses checking for `window.plugins[name]` but I can't figure out where that gets assigned. Is that happening in the caller instead of in this file somewhere? (Or am I just missing it?)

lib/plugin.js
```diff
@@ -130,72 +128,80 @@ const bind = function(name, pluginBind) {
...
+    callback(window.plugins[name])
+    return window.plugins[name]
+  }
+  if (!loadingScripts[name]) { 
+    loadingScripts[name] = (async function() {
+      if (window.plugins[name]) { return window.plugins[name]; }
+      await getScript(`/plugins/${name}/${name}.js`)
+      if (!window.plugins[name]) {
+        await getScript(`/plugins/${name}.js`)
+      }
```
dobbs, line 179: Here's another case where the `async...await` is making this backwards compatibility easier to read.

Tho' it's another place that makes the `window.plugins[name]` part conspicuous. In this context it looks like we expect the plugin author to assign to the global `plugins` and we're checking for that side-effect. (Feeling more convinced this comment is unrelated to this PR, but seems worth thinking about while we're here).

dobbs (commented):
Thanks for working on this, Nick. Thanks especially for separating the conversion to javascript from the conversion to async...await.

I've made several comments to the second of those commits.

lib/plugin.js
```diff
@@ -37,62 +37,60 @@ const loadScript = function(url, options) {
...
       const item = $item.data('item');
-      promise = promise.then(() => new Promise(function(resolve, reject) {
+      try {
         $item.off();
-        return plugin.emit($item.empty(), item, () => resolve());
-      }));
+        await plugin.emit($item.empty(), item)
+      } catch (e) {
+        console.error(e)
+      }
```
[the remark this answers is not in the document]
nrn: Yup, you nailed it. The previous recursive calls built up a promise chain where one broken plugin would stop the chain from progressing. It did have some try/catches and some promise.catches at various stages, but they didn't isolate the plugins from breaking each others ability to emit/bind like this should.

lib/plugin.js
```diff
@@ -37,62 +37,60 @@ const loadScript = function(url, options) {
...
+  for (const itemElem of $items.toArray()) {
     const $item = $(itemElem);
     if (!$item.hasClass('textEditing')) {
       const item = $item.data('item');
-      promise = promise.then(() => new Promise(function(resolve, reject) {
+      try {
         $item.off();
-        return plugin.emit($item.empty(), item, () => resolve());
-      }));
+        await plugin.emit($item.empty(), item)
```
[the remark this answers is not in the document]
nrn: Yeah, the await is a pretty direct replacement for the callback w/ resolve, the main difference is an error thrown in `plugin.emit` would now surface here and get caught. I left the callbacks on all of these functions available but optional, since they are exported, but with promises flowing through all of them now we only have to do the `new Promise(function (resolve...` thing at the edges where this file interacts with external callback functions.

lib/plugin.js
nrn: I did change this a hair, so it doesn't go into the loading scripts mechanism for plugins that are already there. The default plugins all get added to window.plugins in a separate file https://github.com/fedwiki/wiki-client/blob/master/lib/plugins.coffee#L6-L12, and then non-builtin plugins either call `wiki.registerPlugin` from `plugin.registerPlugin` on the last line of this file, or they simply add themselves to `window.plugins` manually.

So the early checks in this file are looking for plugins that have been previously loaded, and the later checks are testing if a plugin has been successfully loaded by seeing if it added itself to `window.plugins`, which also means a plugin can't do anything asynchronous before adding itself to the plugins object or this code thinks that it didn't load successfully. (this part isn't a change, but worth looking at after using import to load plugins lands)

lib/plugin.js
nrn: Yeah, that is exactly what has been happening, this change just makes it easier to understand :sweat_smile: I don't think there is anything we can do to improve it here yet, but this change combined w/ import to load plugins would make it very easy to add an alternate way to register plugins using es6 module exports.

paul90:
There probably needs to be some memory that a plugin has not been found. Each load attempt looks in two places, `/plugins/chess/chess.js` and `/plugins/chess.js`, and this is done twice. Looks as if the second place goes back to a change in plugin location 12 years ago - see [SFW commit](https://github.com/paul90/Smallest-Federated-Wiki/commit/3372b5082e52d10074653c1304abe9d9120f0941).

`plugin.get()` is used in two? places to load the plugin - line 84, in `renderFrom()`, and line 267, in `plugin.emit()`. Once in the render, and then in the bind?

paul90:
Toggling the `wiki` button, on or off, results in content following the item with the failed plugin to get duplicated. The number of times seems to be dependent on the number of pages in the lineup to the left of the page with the failed plugin! Very strange.

Screenshot 2024-07-13 at 11 53 39

This behaviour previously existed.

nrn:
Yeah, I think those will both be easier fixes after this change, but I was trying not to do too much in the one pr. Trying to load the code multiple times may even just be fixed by `import()`, but I saw a potential quick fix for it while I was working on the rest of this I'll add as a suggestion.

lib/plugin.js
```diff
@@ -0,0 +1,287 @@
...
+    await new Promise(function(resolve, reject) {
+      loadScript(url)
+        .done(function() {
+          scripts.push(url);
+          return resolve();
+        }).fail(function(_jqXHR, _textStatus, err) {
+          console.log('getScript: Failed to load:', url, err);
+          return resolve();
+      });
+    })
```
nrn, line 54:
```suggestion
    await new Promise(function(resolve, reject) {
      loadScript(url)
        .done(function() {
          scripts.push(url);
          return resolve();
        }).fail(function(_jqXHR, _textStatus, err) {
          scripts.push(url);
          console.log('getScript: Failed to load:', url, err);
          return resolve();
      });
    })
```

nrn: So, surprisingly this seems to fix both problems, but I don't understand how exactly.

paul90 (approved):
Looks good - `scripts` is a list of the plugins that have been seen. Allows the loading to not repeat trying to load plugins that are not available.

lib/plugin.js
paul90: Looks good - scripts is a list of the plugins that have been seen. Allows the loading to not repeat trying to load plugins that are not available.

WardCunningham (approved):
Thanks to Nick and Paul for walking me through this step by step.

paul90:
As mentioned in the call earlier this week - I have created a branch with`git mv plugin.coffee plugin.js` as the first step. This helps retain the history of the decaffeinate step. A pre-release is available as `wiki-client@0.30.0-rc.1`

paul90 (approved):

paul90:
The force push was to add `git mv` as the first commit, and to convince this PR has been merged.

PR #322, opened 2024-08-25
Decaffeinate -- ESLint and references to wiki module

SamAxe:
This isn't really meant for pulling into this branch yet, only for inspiration if it's useful for you.  

* The first commit adds eslint to grunt `grunt eslint` 
* The second adds module references to wiki.
* The third is starting to replace grunt and browserify with yarn and esbuild (not tested much yet)
* The fifth commit attempts to convert modules from CJS to ES6 style (again, no testing in this branch, just me and my exploration so far, in fact, I learned a lot as I was doing it, so I think there are some issues as it currently is.)


paul90:
Thank you for the work you did here - some very useful ideas here that remain to be done including linting/formatting and update to ESM style.

PR #323, opened 2024-09-02
Extract from currentItem, not item

paul90:
A fix for incorrectly using the non-existent `item`, rather than `currentItem` in `extractPageText()`.

Causes the re-index after editing audio,frame or video items to fail and the client's index to fail.


paul90:
Published as + wiki-client@0.30.1

PR #325, opened 2024-09-09
Remove underscore from the production client.

nrn:
The replacements for the _ functions aren't as general as what was there, but they all seem to suit our purposes.

Underscore is still used in the tests.

### DIFF
--- a/lib/factory.js
+++ b/lib/factory.js
@@ -276,17 +276,17 @@ var arrayToJson = function(array) {
   const cols = array.shift();
   const rowToObject = function(row) {
     const obj = {};
-    for (var [k, v] of Array.from(_.zip(cols, row))) {
+    row.forEach(function (v, idx) {
+      let k = cols[idx]
       if ((v != null) && (v.match(/\S/)) && (v !== 'NULL')) { obj[k] = v; }
-    }
+    })
     return obj;
   };
-  return (() => {
-    const result = [];
-    for (var row of Array.from(array)) {       result.push(rowToObject(row));
-    }
-    return result;
-  })();
+  const result = [];
+  for (var row of array) {
+    result.push(rowToObject(row));
+  }
+  return result;
 };
 
 
--- a/lib/refresh.js
+++ b/lib/refresh.js
@@ -97,7 +97,7 @@ const handleDrop = function(evt, ui, originalIndex, originalOrder) {
 
   if (moveWithinPage) {
     const order = getStoryItemOrder($item.parents('.story:first'));
-    if (arrayShallowEqual(order, originalOrder)) {
+    if (JSON.stringify(order) !==  JSON.stringify(originalOrder)) {
       $('.shadow-copy').remove();
       $item.empty();
       index = $(".item").index($item);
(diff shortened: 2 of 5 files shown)


lib/factory.js
```diff
@@ -276,17 +276,17 @@ var arrayToJson = function(array) {
   const cols = array.shift();
   const rowToObject = function(row) {
     const obj = {};
-    for (var [k, v] of Array.from(_.zip(cols, row))) {
```
WardCunningham, line 279: I wonder, how does _ get defined here since there is no require('underscore') as there is in the other files in this commit?

lib/refresh.js
```diff
@@ -590,4 +588,19 @@ const cycle = function($page) {
   return promise;
 };
 
+function arrayShallowEqual (a, b) {
```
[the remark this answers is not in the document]
nrn: Hmm, I can think of a few cheeky tricks because we know they are both arrays of strings, like `a.join(' ') === b.join(' ')` would work as long as we are sure the items don't contain spaces, but I'm not sure it is an improvement.

lib/factory.js
nrn: Oh, good catch! Looks like we are also injecting it as a script tag on the page, which then causes it to get added as a global. Not sure if plugins depend on that 🙈 This PR at least gets it out of the client bundle/build process.

lib/refresh.js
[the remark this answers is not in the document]
WardCunningham: Interesting alternative. Could join on some odd character like the ASCII Group Separator.

This looked like code that was trying to decide if dragging a page was a rearrangement that required further action. I think the only way to do this reliably in the presence of duplicated pages is to look up the code that represents the panel that holds a page and compare them. I will look into this as I find time today.

lib/refresh.js
```diff
@@ -99,7 +97,7 @@ const handleDrop = function(evt, ui, originalIndex, originalOrder) {
 
   if (moveWithinPage) {
     const order = getStoryItemOrder($item.parents('.story:first'));
-    if (!_.isEqual(order, originalOrder)) {
+    if (arrayShallowEqual(order, originalOrder)) {
```
paul90, line 100: Missing `!`?

WardCunningham: Paul and I added some console.log to confirm that the lists are lists of strings. This from decaffeinate branch without your change.

image

nrn: That is a very good question Paul, I don't know why it seemed to work like this :sweat_smile:

nrn: Oh it definitely does not work, I expected to see a problem while dropping, not having to refresh.

WardCunningham: I recommend we apply toString() to these arrays and then test for string inequality.
image
With Nick's permission I will push this change to his branch.

nrn: Whoops, I just pushed a fix using `JSON.stringify` which seemed to be working well

nrn: It's pretty niche, but `.toString()` could get tripped up if someone generated pages with a `,` in the item ids, but `JSON.stringify()` shouldn't
```
> JSON.stringify([ 1, 2, 3, undefined, null, "4,5"])
'[1,2,3,null,null,"4,5"]'
> [ 1, 2, 3, undefined, null, "4,5"].toString()
'1,2,3,,,4,5'
> 
```

PR #326, opened 2024-09-18
Another DS205 down

paul90:
Not really so complex. Just made so by decaffeinate adding in a return when none was needed.

The standard `git diff` also makes it harder to see what is happening. So here is difftastic view of what I've done.
Screenshot 2024-09-18 at 06 59 23

This section of code is adding highlighting to the search results, all DOM side effects, nothing to return to the caller.


nrn (approved):
Looks good!

PR #327, opened 2024-10-25
display flex for native scrolling

WardCunningham:
@baldurbjarnason has guided me through this modest improvement styling the lineup using flex for native scrolling. This is expected to work better on mobile. (Test this before merging.)

Note: this and related work is more about investing in the longevity of our codebase than adding mobile as a feature. I once referred to this as "chronological scaling" before we had a clear notion of what that would require. [post](http://code.fed.wiki/chronological-scaling.html)

PR #328, opened 2024-11-03
display flex for native scrolling

paul90:
Replacing #327, which used the coffeescript, now legacy, branch as a starting point.

(cherry picked from commit fe55e8264bb73f4427c8ad60792d25b45ecc9585)

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -42,10 +42,14 @@ body {
   left: 0;
   right: 0;
   bottom: 0;
-  position: absolute;
+/*  position: absolute;*/
   bottom: 60px;
   margin: 0;
-  width: 10000%; }
+/*  width: 10000%; */
+  display: flex;
+  overflow-x: auto;
+  height: calc(100vh - 40px);
+}
 
 footer {
   border-top: 1px solid #3d3c43;
@@ -391,8 +395,9 @@ p.readout {
   margin: 8px;
   /* padding: 0 31px; */
   width: 490px;
+  flex: 0 0 490px;
   background-color: white;
-  height: 100%;
+/*  height: 100%;  */
   overflow: auto;
   box-shadow: 2px 1px 4px rgba(0, 0, 0, 0.2);
   outline: none; }
--- a/lib/active.js
+++ b/lib/active.js
@@ -4,45 +4,49 @@
 let active;
 module.exports = (active = {});
 
-active.scrollContainer = undefined;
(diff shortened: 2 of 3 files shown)


paul90:
Comments from #327 

https://github.com/fedwiki/wiki-client/pull/327#issue-2614768231 (@WardCunningham )
> @baldurbjarnason has guided me through this modest improvement styling the lineup using flex for native scrolling. This is expected to work better on mobile. (Test this before merging.)
>
>Note: this and related work is more about investing in the longevity of our codebase than adding mobile as a feature. I once referred to this as "chronological scaling" before we had a clear notion of what that would require. [post](http://code.fed.wiki/chronological-scaling.html)

https://github.com/fedwiki/wiki-client/pull/327#issuecomment-2439483784 (@paul90)
> Only thing I've spotted so far is that the horizontal scrollbar on the lineup needs hiding, and page being opened no longer is scrolled into view. No auto scrolling the lineup to bring the new page into view ~~is most likely a fault in the JS conversion!~~, looks to be something jQuery related that get broken here.

https://github.com/fedwiki/wiki-client/pull/327#issuecomment-2439742726 (@WardCunningham)
> @paul90 recommended this simple fix, [bfa2e8b](https://github.com/fedwiki/wiki-client/pull/327/commits/bfa2e8b5f985e4b74e3aefde94bb643ba7879ac8), and adds, "if we were not moving away from browserify it could be fixed by adding [browserify-optional](https://github.com/devongovett/browserify-optional) into the mix."
>
> Rendering on older iPad.
IMG_0053
>
> Rendered on old iPhone.
IMG_9404

https://github.com/fedwiki/wiki-client/pull/327#issuecomment-2451079169 (@WardCunningham)
> Direct positioning within the lineup happens when a specific page is made active. This is handled by code in [active.coffee](https://github.com/fedwiki/wiki-client/blob/master/lib/active.coffee). Changes here have thrown off these calculations.
>
>Various places in the code will expose a particular page and make it active with idiomatic code that looks like:
>
>`active.set($('.page').last())`
>
> We see calls like this in these modules:
> - factory.coffee -- once
> - link.coffee -- twice
> - state.coffee -- once
> - search.coffee -- unused, now obsolete by pop-up?
> - legacy.coffee -- seven times, in various global event handlers

paul90:
Do we want to suppress the horizontal scrollbar on the lineup? Or is that needed for mobile/tablet.

PR #329, opened 2024-11-03
fix arrow keys and local storage

WardCunningham:
I don't know what I was thinking when I replaced a convoluted switch statement with a malformed table lookup.
This expression has been shown to work.

### DIFF
--- a/lib/legacy.js
+++ b/lib/legacy.js
@@ -51,10 +51,9 @@ $(function() {
 
   const LEFTARROW = 37;
   const RIGHTARROW = 39;
-  const dirs = {LEFTARROW:-1, RIGHTARROW:+1};
 
   $(document).on("keydown", function(event) {
-    const direction = dirs[event.which];
+    const direction = event.which == LEFTARROW ? -1 : event.which == RIGHTARROW ? 1 : null
     if (direction && !$(event.target).is(":input")) {
       const pages = $('.page');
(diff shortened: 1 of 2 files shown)


paul90 (approved):

paul90:
Could have just reverted to the original, from the decaffeinate script:
``` javascript
    const direction = (() => { switch (event.which) {
      case LEFTARROW: return -1;
      case RIGHTARROW: return +1;
    } })();
```

lib/legacy.js
```diff
@@ -51,10 +51,9 @@ $(function() {
 
   const LEFTARROW = 37;
   const RIGHTARROW = 39;
-  const dirs = {LEFTARROW:-1, RIGHTARROW:+1};
```
paul90, line 54: This should have been
``` javascript
const dirs = {[LEFTARROW]:-1, [RIGHTARROW]:+1};
```

WardCunningham: Looking a few lines further down I see more complex case analysis and conclude this is really two distinct event handlers struggling to be one.

WardCunningham:
The return was important because it was an early exit, not because anything of value was being returned. There is a structured-programming rule, single entry with single exit, that would have had us coding that as an if-then-else then no return would have been required. The early exit return should be coded as a separate statement to make it clear that the return is important for control flow.

PR #336, opened 2025-02-26
Update future.js

Bortseb:
Change the logic of when the alternate lineup is suggested over http. It's currently suggested only when there were alternate pages to offer that could be reached, and not when none of those pages could be reached. This seems backwards.

also, Fix what seems to be a typo.


Bortseb:
Or maybe this the alternate lineup should just always be shown when in a secure context? This is what I am shown when there were other pages to offer that were "reachable", but even then the altenate lineup over http still seems useful...

image

PR #342, opened 2025-05-27
ghost rename

WardCunningham:
This is a rewrite of the incomplete implementation decaffeinating the work in progress #291

### DIFF
--- a/lib/legacy.js
+++ b/lib/legacy.js
@@ -311,6 +311,12 @@ from an export file dated ${file.lastModifiedDate}.\
         if ($page.data('rev') != null) {
           $page.find('.revision').remove()
         }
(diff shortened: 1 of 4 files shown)


WardCunningham:
Nick and I have revised the text handling for renamed ghost pages.
- configure contenteditable to use plaintext-only
- replace any string of whitespace (especially unexpected newlines) with a single space
- apply replacements before detecting a title edit

WardCunningham:
We have considered that the edits we apply to a title do not show up until the page is saved. We think this is sufficient because it does allow the author to revise their edits before fork even if this is unnecessary. But most importantly careless spacing mistakes are not recorded in the saved json. That means if we should ever choose to do more visual confirmation we won't have to worry about unwanted saved data.

PR #344, opened 2025-06-27
use narrow page on narrow devices

WardCunningham:
Small adjustments view in the chrome mobile device simulator.

image

This specific complaint is from Thompson who writes,
> When I send pages to others with the likelihood they will be read in their phone, I do so being aware that since the last major update, they don’t render well as the text width of the page is larger than the screen size, requiring some annoying side to side scrolling… I wonder if I’m the only one seeing this issue…
http://thompson.forage.ustawi.wiki/view/vagus-nerve

This css is from Baldur Bjarnason who suggested it as a quick fix without getting into page formatting theory.

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -260,7 +260,7 @@ div.image {
 /* end image plugin editor */
 
 .backlinks {
-  width: 420px;
+  /*width: 420px;*/
   margin-top: 2px;
   clear: both;
   background-color: #eeeeee;
@@ -294,7 +294,7 @@ div.image {
 }
 
 .journal.ui-draggable {
-  width: 420px;
+  /*width: 420px;*/
 }
 
 .action.fork {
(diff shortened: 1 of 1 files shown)


WardCunningham:
image

WardCunningham:
Commit [60adb2e](https://github.com/fedwiki/wiki-client/pull/344/commits/60adb2ec1149d01e3c07c58bbb31472a44a6ce7b) addresses the width and overflow adjustments discussed yesterday. https://github.com/fedwiki/wiki-client/pull/344#issuecomment-3036976313

Coding seems a bit safer in this version.

We know that some plugins, such as the Video plugin, don't play by the same rules and will induce additional scrolling in the mobile configuration. Paul, Eric and I have coded these mods together and don't think misbehaving plugins should stop this from going forward.

paul90 (approved):
As noted elsewhere, there a few plugins that have the 420px width, but this looks good.

paul90:
Published as + wiki-client@0.31.3

PR #347, opened 2025-08-21
adding support for relative external links

paul90:
As mentioned on 20 Aug, adding support for using relative URLs in an external link. This removes the need for an external link to start with a protocol, though relative URLs must start with a `/`. 

Replace `rel="nofollow"`, as it has no real meaning, with `rel="noopener"`, so the new tab/window will not share browsing context. 

Use non-capturing groups for thing being matched (protocol), but we don't use in the replacement function, and remove them from the function's signature.

*There is a remote risk that there is text match somewhere that will match as an external link, when it shouldn't. Probably a risk worth taking? Thoughts.*

### DIFF
--- a/lib/resolve.js
+++ b/lib/resolve.js
@@ -47,12 +47,10 @@ resolve.resolveLinks = function (string, sanitize = escape) {
     }
   }
 
-  const external = (match, href, rest) => {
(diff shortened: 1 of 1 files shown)


WardCunningham (approved):
Thanks for the zoom walkthrough.

PR #348, opened 2025-09-06
async fixes by Ward, Nick, and Paul

WardCunningham:
We also deprecate callback on emit which exists only in the Method plugin which Ward will upgrade.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -138,6 +138,7 @@ const bind = function (name, pluginBind) {
 
 plugin.wrap = wrap
 function wrap(name, p) {
+  if (p.emit.length > 2) console.warn(`Trouble with ${name}: plugin interface no longer accepts callback functions.`)
   p.emit = emit(p.emit)
   p.bind = bind(name, p.bind)
   return p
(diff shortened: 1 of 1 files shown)


WardCunningham:
Any other plugin expecting callbacks will be warned.

image

I'm unsure how Method synchronized. We did test that the bottles of beer computation still works.
http://start.fed.wiki/seventh-floor.html

lib/plugin.js
```diff
@@ -230,13 +231,9 @@ plugin.emit = async function (div, item, done = () => {}) {
...
       dialog.showModal()
       $('.close').on('click', () => dialog.close())
-      $('.retry').on('click', function () {
-        if (script.emit.length > 2) {
-          script.emit(div, item, () => done())
-        } else {
-          script.emit(div, item)
-          done()
-        }
+      $('.retry').on('click', async function () {
```
dobbs, line 234: This is maybe not that important, but I think both the `close` and `retry` handlers would be improved by providing the dialog element as context like so:
``` js
$('.close', dialog).on('click', ...)
$('retry', dialog).on('click', ...) 
```
I think this would only matter if there were multiple errors showing in a lineup, which is probably uncommon.

WardCunningham: I'm unable to find any jQuery documentation that explains what this second parameter is or means. This is as close as I come but this isn't what is being suggested here, I think.

https://api.jquery.com/context/

How would this second parameter be used? Are there further modifications required to make it work?

Thanks. Sorry to be so slow responding.

dobbs: `$('.close').on('click', ...)` adds a click handler that applies to every element in the entire `document` with the `close` class. By specifying the second parameter it reduces the scope from `document` to just the specific `dialog` element.

dobbs: Because we don't have many place in the code where `retry` or `close` classes are added, this is a pretty minor issue, I think.

PR #349, opened 2025-09-18
Fix IPv6 handling in siteAdapter (bracketed hosts, proxy path, cache keys)

RalfBarkow:
This PR should fix longstanding issues when running a wiki with IPv6 neighbors (e.g. `http://[::1]:4000`) on Ubuntu and other systems with IPv6 enabled.

#### Motivation

* Without brackets, raw `::1:4000` hosts caused `Invalid URL` errors in browsers and `curl`.
* Proxy paths were unsafe because raw IPv6 literals were embedded directly in `/proxy/...` without encoding.
* Localhost detection only matched `"localhost"`, not `::1` or `[::1]`.
* Cache keys in `localForage` (`routes`, `withCredentials`) were inconsistent: `::1:3000` vs `[::1]:3000`.
* Some functions (`getURL`, `getDirectURL`, journal fixups) did brittle string surgery with `substring(7)`.

#### Changes

* Add helpers:

  * `normalizeSite`, `siteKey` → ensure canonical bracketed form
  * `httpOriginFor`, `httpsOriginFor`, `proxyPathFor` (encoded)
  * `isLoopbackSite` detection (supports `::1`, `[::1]`, IPv4, localhost)
* Update `findAdapterQ`:

  * Prefer direct `http://[::1]:port/...` for loopback
  * Use proxy path for HTTPS origins
  * Fallback to HTTPS direct if proxy fails
* Replace protocol-relative URLs (`//host/...`) with explicit schemes.
* Replace string hacks with base reconstruction logic.
* Canonicalize keys in `routeStore` and `withCredsStore`.
* Clean up `findAdapter` (proper arrow body).
* Use template strings for flag replacement consistently.

#### Testing

* Local Test Farm setup with two wiki instances (ports 3000, 4000).
* On Ubuntu with IPv6:

  * `curl http://[::1]:4000/favicon.png` returns 200
  * `curl http://localhost:3000/proxy/%5B::1%5D:4000/favicon.png` returns 200
* Browser tests:

  * On HTTP pages → direct bracketed URLs
  * On HTTPS pages → proxy paths with encoded IPv6
* Verified flag replacement and journal/fork links resolve correctly.

#### Related

* Closes issue(s): IPv6 proxy/localhost troubleshooting reported on Ubuntu.

---

### Checklist

* [x] Bracket IPv6 hosts, preserve ports
* [x] Encode proxy path segments safely
* [x] Canonicalize cache keys
* [x] Fix broken `findAdapter` syntax
* [x] Tested on Ubuntu with IPv6 loopback


lib/siteAdapter.js
```diff
@@ -7,6 +7,36 @@ const localForage = require('localforage')
 
 module.exports = siteAdapter = {}
 
+// IPv6/localhost-safe URL helpers -------------------------------------------
+const isLoopbackHost = h =>
+  h === 'localhost' || h === '127.0.0.1' || h === '::1' || h === '[::1]';
```
paul90: This needs to cater for localhost subdomains.

Which is what `if (site.split('.').at(-1).split(':')[0] === 'localhost') {` is part of.

paul90 (commented):
From just an initial cursory glance.

These changes will also have an impact server side, and anywhere wiki names are used. A quick test shows it does not work with a server running in farm mode.

PR #350, opened 2026-04-06
Don't automatically capitalize Display Name in footer

Bortseb:
Set the display name exactly as it is in the owner.json file.

### DIFF
--- a/views/static.html
+++ b/views/static.html
@@ -32,7 +32,7 @@
     <footer>
       <div id='site-owner' class='footer-item'>
         {{#owned}}
-          Site Owned by: <span id='site-owner'>{{ownedBy}}</span></div>
+          By: <span id='site-owner'>{{ownedBy}}</span></div>
         {{/owned}}
       </div>
 
--- a/views/static.html
+++ b/views/static.html
@@ -32,7 +32,7 @@
     <footer>
       <div id='site-owner' class='footer-item'>
         {{#owned}}
-          Site Owned by: <span id='site-owner' style='text-transform:capitalize;'>{{ownedBy}}</span></div>
+          Site Owned by: <span id='site-owner'>{{ownedBy}}</span></div>
         {{/owned}}
       </div>
 


paul90:
Been like that for over 10 years... why you want change after all these years. Though I guess it would be better in a stylesheet.

Bortseb:
Just thought authors might want to have their display name show up how they want it, which isn't necessarily to capitalize everything. Or maybe its more like a fun username or something, rather than a proper noun.

Bortseb:
Some intentionally want everything lower case,  or maybe they have a surname prefix that isn't supposed to be capitalized. for example particles in surnames like "de," "von," or "van" are not capitalized.

Bortseb:
Was also considering proposing "Site Owned by:" just changes to "By:" as the former is really long and takes up a lot of space in the footer, especially when opened in a mobile browser... Also, "ownership" seems a little tangential to the idea of contributing to a creative commons, where everything is licensed CC BY-SA 4.0

WardCunningham:
Do we even need to expose this field? It seems this is between the author and the provider, not automatic public knowledge. If authors want to be known by name they should write it the way they want on their welcome page.

Bortseb:
Ah, also an interesting thought to just remove it entirely :P That would definitely free up space in the footer for mobile-friendly interface... especially if someone has chosen a longer display name...

Though, I guess having it in the footer does give a little more context at all times, if welcome visitors isn't in the lineup... and the domain doesn't give any clues.

Bortseb:
This came up for me because I was adding in a new field on our registration form for keycloak, to let the person choose their display name that gets used...

If display name was eliminated, that would simplify things... Especially since, if someone changes their display name on keycloak, it won't automatically update all the owner files.

Same goes for the username field in the OAuth2 section of the owner file. We are using keycloaks unique IDs per user for the OAuth2 ID field... Butz the owner file also includes the username attribute, which can also change. (Using keycloaks unique ID actually is crucial to enabling usernames to change)

Usernames also don't wind up being public, it's only in the ownerfile... Not sure if username is useful to have their either, since it seems only the ID parameter is used for authentication.

paul90:
> Was also considering proposing "Site Owned by:" just changes to "By:" as the former is really long and takes up a lot of space in the footer, especially when opened in a mobile browser... Also, "ownership" seems a little tangential to the idea of contributing to a creative commons, where everything is licensed CC BY-SA 4.0

I don't see it as tangential, all part of attribution. Without it the wiki would appear the same as an unclaimed one, lacking any indication of ownership.

paul90:
> This came up for me because I was adding in a new field on our registration form for keycloak, to let the person choose their display name that gets used...
> 
> If display name was eliminated, that would simplify things... Especially since, if someone changes their display name on keycloak, it won't automatically update all the owner files.
> 
> Same goes for the username field in the OAuth2 section of the owner file. We are using keycloaks unique IDs per user for the OAuth2 ID field... Butz the owner file also includes the username attribute, which can also change. (Using keycloaks unique ID actually is crucial to enabling usernames to change)
> 
> Usernames also don't wind up being public, it's only in the ownerfile... Not sure if username is useful to have their either, since it seems only the ID parameter is used for authentication.

Remember your situation with keycloak is somewhat unique, you know who everybody is. For a more typical farm that is not likely to be the case. In any case if all you have is a display name, an id in some third-parties identity system, and a few people using the same display name, life might get difficult in the event of an incident.

There is a case for updating the owner file with changed display name, something for the replacement security plugin.

Bortseb:
> > Was also considering proposing "Site Owned by:" just changes to "By:" as the former is really long and takes up a lot of space in the footer, especially when opened in a mobile browser... Also, "ownership" seems a little tangential to the idea of contributing to a creative commons, where everything is licensed CC BY-SA 4.0
> 
> I don't see it as tangential, all part of attribution. Without it the wiki would appear the same as an unclaimed one, lacking any indication of ownership.

Oh, I didn't really mean tangential to mean that I think we shouldn't indicate at all whether a wiki is claimed or unclaimed... I was just searching for a better word to describe the state of the wiki other than "ownership"... even just "By: {display_name}" would be better to serve as the [Byline](https://en.wikipedia.org/wiki/Byline) in my opinion.

I've just always felt that what we are doing with federated wiki is "commoning" and building a commons together that isn't really "owned" by anyone in like a typical capitalist sense.. so just the concept of "ownership" always felt a little off to me, and it appears on every site.

So if someone really cares about their display name/ byline, then they might also have a little bit of an allergic reaction to claiming the site is "owned" by them too...

Haha, sorry if this seems like way too much fuss over some small detail :P

Bortseb:
> > This came up for me because I was adding in a new field on our registration form for keycloak, to let the person choose their display name that gets used...
> > If display name was eliminated, that would simplify things... Especially since, if someone changes their display name on keycloak, it won't automatically update all the owner files.
> > Same goes for the username field in the OAuth2 section of the owner file. We are using keycloaks unique IDs per user for the OAuth2 ID field... Butz the owner file also includes the username attribute, which can also change. (Using keycloaks unique ID actually is crucial to enabling usernames to change)
> > Usernames also don't wind up being public, it's only in the ownerfile... Not sure if username is useful to have their either, since it seems only the ID parameter is used for authentication.
> 
> Remember your situation with keycloak is somewhat unique, you know who everybody is. For a more typical farm that is not likely to be the case. In any case if all you have is a display name, an id in some third-parties identity system, and a few people using the same display name, life might get difficult in the event of an incident.
> 
> There is a case for updating the owner file with changed display name, something for the replacement security plugin.

both friends and passportjs security plugins seems pretty similar to me. Both have a "name" parameter in the owner.json file, and then either a secret or an ID ... (and I guess you're saying the display name is maybe more for the admin to keep track of things? rather than some aspect of a wiki site that an author might want to customize? even though it appears in the footer of their site)

Yes, using keycloak with passportjs is different... even if the only parameter the owner.json file included was the unique ID, we could still look them up in our user database to track down who that person is.... So if the display name was missing, or removed, it wouldn't be as big a deal for us... But even with a "typical" farm, the display name might not be of much use to the admin unless they enforce some sort of rule around what display names can be... as the user has no means to set their display name themselves, other than making a request to an admin on the farm... And likely if they just claimed a site themselves in the traditional way, their display name is just some random thing that was generated for them. It was only somewhat recently that the register plugin enabled any owner to use `register-delegate-site` to create a new site for someone else, and specify their display name at the time of claiming it for them... (though, this only works for friends-security currently. It wouldn't work for passportjs unless the admin wants to go look up the person's unique ID in keycloak)

Though, claiming a site in the typical way still works with the passportjs+keycloak case... the person claiming the site just needs an existing account on keycloak, and then to claim the site they just need to authenticate.

paul90:
Only it is not a byline! It's an indication of ownership of the wiki. Content could, thanks to forking, be by others. "Owner: " would be a better shortening...

Bortseb:
Ah, I guess I wasn't really thinking of it in like a technical/legal way... I guess my creation would have to be licensed under Public Domain or Creative Commons Zero (CC0) for me to properly waive all legal "ownership"?

And maybe byline is also technically the wrong term too, it just seemed like the closest fit for a related concept... Can't the "ownership" just be implied like with a byline? Why must it be so foregrounded/explicit?

If I'm creating something to give away or share as best I can, I stop thinking as if I own that thing... Especially if where it exists is in some digital space that I probably rent from someone else, to use Ward's apartment building metaphor.

Bortseb:
If we must keep a verb in there, I'd prefer created by, or authored by... But my top preference would still be just to keep it short and say "By: {display_name}"

almereyda:
Can it be "Written by", "Owned by", "Authored by" or be made configurable?

I'm all for omitting technical details, such as "Site " for regular users.

A proud statement will be "Wiki by" or "Wiki'd by", lending wikis progressive nature a hand and turning it into a verb, very much like "to google": "to wiki". But we have that term already in the footer and don't need to repeat it.

Just "By:" Is a little too less descriptive, from my angle.

These kinds of strings could also be thought together with a broader

- https://github.com/fedwiki/wiki/issues/139

of things. When UI strings are just some kind of extra file, everyone can replace them as they wish.

PR #351, opened 2026-04-20
Fix build syntax to work on windows

Bortseb:
The build script is using semicolons (;) to chain commands, but on Windows, npm needs && to properly chain commands. The semicolon is being treated as part of the script name.


paul90:
That is not how `&&` works.

PR #353, opened 2026-06-18
Add horizontal scroll snapping when in mobile mode

Bortseb:
Add horizontal scroll snapping when in mobile mode

You can test this change out on this site:
https://mobile-scroll-snap.ux.aolc.cc/

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -528,6 +528,11 @@ textarea {
     flex: 0 0 100vw;
     font-size: 0.87rem;
     overflow-x: unset;
+    scroll-snap-align: center;
+  }
+
+  .main {
+    scroll-snap-type: x mandatory;
   }
 }
 


WardCunningham:
Screen shots from my phone held in landscape and portrait orientations. Markdown's unequal image scaling makes it hard to compare. But important thing to notice is that the page heading folds in one case but not the other. I assume this is a consequence of the font scaling inherited from the existing media specific css and not a result of this request.

I notice that the scaling was applied by my own commit a year ago. https://github.com/fedwiki/wiki-client/commit/c31319f2bf816bb0965c633ddac21229bf3d8491

Suggest we accept this pr as-is and consider fine-tuning size in a separate investigation.

## landscape orientation
IMG_0692

## portrait orientation
IMG_0690

Bortseb:
That landscape view is very strange...no idea how the page width is getting so large... this doesn't happen on my mobile device

PR #354, opened 2026-06-20
Clean up default style.css

Bortseb:
Remove dead and redundant rules from the default wiki stylesheet.

- Remove empty .journal.ui-draggable rule with commented-out width
- Remove unused image.remote selector (remote icons use img.remote in markup)
- Simplify mobile .page styles: drop redundant margin, padding, and overflow-x overrides; use standard 0.875rem font size

You can test these changes out at:

https://clean.aolc.cc/

### DIFF
--- a/client/style/style.css
+++ b/client/style/style.css
@@ -42,14 +42,7 @@ body {
 }
 
 .main {
-  top: 0;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  /*  position: absolute;*/
-  bottom: 60px;
   margin: 0;
-  /*  width: 10000%; */
   display: flex;
   overflow-x: auto;
   height: calc(100vh - 40px);
@@ -62,6 +55,7 @@ footer {
   position: fixed;
   left: 0;
   right: 0;
+  bottom: 0;
   height: 20px;
   padding: 10px;
   font-size: 80%;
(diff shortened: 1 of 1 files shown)


Bortseb:
There were duplicate .footer and .factory rules, consolidated

Bortseb:
ok, i had a to fix a few things here... 

Here is better justification for each change:

| Change | Why it’s safe |
|---|---|
| Drop `.main` `top` / `left` / `right` / `bottom` | `.main` is not positioned (`position: absolute` is commented out). Inset properties have no effect. |
| Move `bottom: 0` into the main `footer` block | Same computed style as the old separate `footer { bottom: 0; }` rule — keeps the fixed chrome bar pinned without a duplicate block. |
| Delete `.twins:hover, .journal:hover, .footer:hover { opacity: 1; }` | No-op: the base `.twins, .journal, .footer` rule already sets `opacity: 1`. |
| Delete empty `.journal.ui-draggable` | Only contained a commented-out `width`; no live declarations. |
| Delete `image.remote` | Wrong selector. Markup uses `img.remote`, already covered by `img.remote, .neighbor img`. |
| Merge the two `.factory` blocks | one combined rule is equivalent. |
| Mobile: drop `width: 1%`, `margin: 8px`, `padding: 0`, `overflow-x: unset` | `flex: 0 0 100vw` already sizes the page; base `.page` already has `margin: 8px` and no padding; `overflow: auto` remains. |
| Mobile: `0.87rem` → `0.875rem` | Standard 14/16 rem; tiny intentional tidy. |

PR #356, opened 2026-08-03
Enable video drops onto factory from links or files

Bortseb:
Support dropping HTML5 video files and URLs onto factory

Upload mp4/webm via the assets plugin to "plugins/video/" and recognize direct .mp4/.webm URL drops, creating video items with the correct subtype instead of punting.

Consider adding an asset item onto about-video-plugin page showing all videos in "plugins/video/"

### DIFF
--- a/lib/drop.js
+++ b/lib/drop.js
@@ -54,6 +54,25 @@ const isSvg = function (url) {
   return null
 }
 
+// HTML5 video subtype keyed by file extension / mime subtype.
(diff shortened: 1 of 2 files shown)


paul90:
Appears to combine two probable issues, dropping a url that points to a HTML5 video file, and dropping the file itself. There has been no real discussion about either. While was something about using the video plugin rather using a video tag in the HTML plugin, but there was nothing about dropping the URL into the factory. And no discussion about if the video plugin should take on asset management of video artifacts.

PR #357, opened 2026-08-06
Persist wiki edit mode and sync it on login/logout

Bortseb:
## Summary
- Persist the footer **wiki ✔︎** toggle in `localStorage` (`wikiEditEnable`) so a manual choice survives reload.
- When no preference is saved, default edit **on** for authenticated desktop and **off** for logged-out users or narrow viewports (`max-width: 490px`, same breakpoint as page CSS).
- On login/logout (security footer `#security` update), match classic wiki behavior: **login → edit on** (any viewport), **logout → edit off**, including when a prior preference was saved.
- Only react when `isAuthenticated` actually changes, so the security plugin’s initial footer paint does not overwrite a saved preference.

### DIFF
--- a/lib/legacy.js
+++ b/lib/legacy.js
@@ -458,21 +458,37 @@ Installed plugins offer these utility pages:\
       })
     })
 
+  const setEditState = function (editState) {
+    $('.editEnable').toggle(!!editState)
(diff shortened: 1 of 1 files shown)


paul90:
What is the issue this is trying to fix? Something that probably needs discussion in an issue.

PR #358, opened 2026-08-07
Keep incremental search open after results are added to lineup, and improve result keyboard UX

Bortseb:
## Summary
- Stop clearing `.incremental-search` when a page becomes active (fixes mech/page opens dismissing search mid-query).
- Opening a result keeps focus on that link so you can can keep opening more results
- Tab cycles through the search input and result links and doesn't escape into the rest of the wiki; Escape or click-away dismisses.
- Clicking the search field again re-shows results if the query is still there (e.g. after Escape).

This feature was inspired by watching the mech add results to the lineup on this site:
http://jargon.fed.wiki/

I found myself wanting to search for the pages that flashed in front of me, and realized the search results were quickly closed as soon as the mech added a new page to the lineup, because it was made active and stole focus.

### DIFF
--- a/lib/active.js
+++ b/lib/active.js
@@ -50,7 +50,6 @@ function scrollTo($page) {
 
 active.set = function ($page, noScroll) {
   if ($page == null) return
-  $('.incremental-search').remove()
   $page = $($page)
   $('.active').removeClass('active')
   $page.addClass('active')
--- a/lib/search.js
+++ b/lib/search.js
@@ -68,7 +68,10 @@ const createSearch = function ({ neighborhood }) {
           // ensure that name is a string (using string interpolation)
           name = `${name}`
           pageHandler.context = $(e.target).attr('title').split(' => ')
-          return finishClick(e, name)
+          finishClick(e, name)
+          // keep focus on clicked result so more results can be opened
+          e.target.focus()
+          return false
         })
         .on('click', 'img.remote', function (e) {
           // expand to handle click on temporary flag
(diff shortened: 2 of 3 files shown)


Bortseb:
One small addition in active.set’s scroll path: while .incremental-search is open on a mobile viewport (max-width: 490px), skip scrollIntoView. Opening a hit still activates the page, but on a phone that smooth scroll fights the soft keyboard / results overlay. Desktop behavior is unchanged — the opened page still scrolls into view.

paul90:
Adds nothing to the core client that is not already there. What you describe sounds more like a problem either with how you are using mech, or in mech itself. You can prevent mech for running by itself by adding a click.

Bortseb:
anytime a page becomes active it closes the search results, the mech example I gave was just how I discovered it, because it was making pages active every time a tick happened, so search was rendered useless until the mech completed its ticks.

The bullets  above describe what was added.

- In core: When you click a result, the search overlay closes... you aren't enabled to easily open more than one result. (Similar on mobile, when a result is clicked, that page becomes active and fights with the mobile results and keyboard staying open)
- in core: if you are tabbing through results, it will escape out and tab through the rest of the links in the lineup/interface.
- yes the last bullet was already a feature it seems, nothing new.

Maybe you just don't think any of these new features are worth adding? So much so that they are considered to be nothing?

paul90:
simple solution if you want more than one result simply press return and get the ghost page of search results.

